PR #78, opened 2024-12-02
Normalize all line endings to Unix on the way in

### LINKED ISSUE #74: Normalize line endings in multiline strings
See also https://github.com/posit-dev/air/pull/69#issuecomment-2502473548

We have this CI failures on Windows https://github.com/posit-dev/air/actions/runs/12042018483/job/33574939950?pr=69

With this test

```
---- formatter::r::specs::r::value::string_value_r stdout ----
thread 'formatter::r::specs::r::value::string_value_r' panicked at C:\Users\runneradmin\.cargo\git\checkouts\biome-d49ce8898b420a50\2648fa4\crates\biome_formatter\src\builders.rs:394:5:
The content 'r"("some raw string
business")"' contains an unsupported '\r' line terminator character but text must only use line feeds '\n' as line separator. Use '\n' instead of '\r' and '\r\n' to insert a line break in strings.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Which formats a file containing this string

```r
r"("some raw string
business")"
```

Note that this is a _multiline string_. The raw string there isn't critical to the reprex, it's just what we happened to have there.

This fails because:
- On windows, git checks the file out with CRLF line endings. Note this puts a `\r\n` inside the string contents.
- The formatter fails inside `located_token_text()` for a string because of this [no-\r assertion](https://github.com/biomejs/biome/blob/cd1c8ec4249e8df8d221393586d664537c9fddb2/crates/biome_formatter/src/builders.rs#L365)

Both biome and ruff use this assertion, so I imagine it is for a good reason.

The way both of them handle this is by using variants of `normalize_string()` here:
https://github.com/biomejs/biome/blob/cd1c8ec4249e8df8d221393586d664537c9fddb2/crates/biome_formatter/src/token/string.rs#L48

Note that the docs say one of the things it does is replace `\r\n` with just `\n`.

They call this at `fmt()` time for a string node, before it actually drops through to `located_token_text()`, so they get a chance to "fix up" the string here. We could probably create our own version of this and strip out everything except the `\r\n` conversion, until we decide to do more normalization.

## Why do they even do this?

I imagine the idea is that typically the line ending doesn't really matter, it becomes part of the Trivia set, but is not part of an actual node. But here, the line ending gets parsed into the actual string contents. I assume they want to ensure that the actual nodes of the tree look the same across OSes regardless of line endings, and standardizing on `\n` inside multiline strings is the way to achieve that.

DavisVaughan:
Closes #74


.gitattributes
```diff
@@ -1 +1,4 @@
 * text=auto eol=lf
+
+# Windows specific test files where we need CRLF endings
+crates/air_r_formatter/tests/specs/r/crlf/*.R text eol=crlf
```
DavisVaughan, line 4: This sets it so that all files are _checked out_ with LF endings except this one directory, which is forced to have CRLF endings (even on Mac, and particularly on CI)

crates/air/src/commands/format.rs
```diff
@@ -29,16 +30,30 @@ fn _format_dir(path: &PathBuf) -> anyhow::Result<ExitStatus> {
...
     }
 
-    let formatter_options = RFormatOptions::default();
+    // TODO: Respect user specified `LineEnding` option too, not just inferred line endings
+    let line_ending = match line_ending {
+        LineEnding::Lf => biome_formatter::LineEnding::Lf,
+        LineEnding::Crlf => biome_formatter::LineEnding::Crlf,
+    };
+
+    let formatter_options = RFormatOptions::default().with_line_ending(line_ending);
```
DavisVaughan, line 56: The idea is that the cli should
- infer the line endings from the contents
- normalize to unix before parsing/formatting
- convert back to inferred line endings on the way out (eventually respecting a forced `LineEnding` option if the user set one)

crates/air_formatter_test/src/spec.rs
```diff
@@ -28,6 +28,9 @@ impl<'a> SpecTestFile<'a> {
 
         let input_code = std::fs::read_to_string(input_file).unwrap();
 
+        // Normalize to Unix line endings
+        let input_code = line_ending::normalize(input_code);
+
```
DavisVaughan, line 33: We don't have a way to set options on a per test basis right now, but eventually it will be nice to have a test where we force the format output to be CRLF to ensure it is working right

crates/line_ending/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+//! We maintain the invariant that all internal strings use `\n` as line separator.
+//! This module does line ending conversion and detection (so that we can
+//! convert back to `\r\n` on the way out as needed).
+
+use std::sync::LazyLock;
+
+use memchr::memmem;
+
+static FINDER: LazyLock<memmem::Finder> = LazyLock::new(|| memmem::Finder::new(b"\r\n"));
```
DavisVaughan, line 9: One of the benefits of `Finder` is that you can construct one up front and reuse it, like with regexes
https://github.com/astral-sh/uv/blob/81569c47bfa91b24ff0712baf1c001ef9604676e/crates/uv-scripts/src/lib.rs#L17

crates/line_ending/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum LineEnding {
+    ///  Line Feed only (\n), common on Linux and macOS as well as inside git repos
+    Lf,
+
+    /// Carriage Return + Line Feed characters (\r\n), common on Windows
+    Crlf,
```
DavisVaughan, line 17: I switched to the convention that `biome_formatter::LineEnding` uses so it maps over nicely

crates/line_ending/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+/// safe even if a panic occurs.
+///
+/// # Source
+///
+/// ---
+/// authors = ["rust-analyzer team"]
+/// license = "MIT OR Apache-2.0"
+/// origin = "https://github.com/rust-lang/rust-analyzer/blob/master/crates/rust-analyzer/src/line_index.rs"
+/// ---
+pub fn normalize(x: String) -> String {
```
DavisVaughan, line 43: I feel like it is more flexible to move towards providing attribution in doc comments

- It lets us structure the folders however we want, without needing a `rust-analyzer` folder
- It allows us to change function names and whatnot while still pointing to the original source

DavisVaughan:
I feel pretty good about this one, we can iterate as needed!

PR #80, opened 2024-12-03
Remove the global `AUXILIARY_EVENT_TX` and use `tracing`

### LINKED ISSUE #79: `test_format_minimal_diff` is not reliable
We occasionally see this in CI without any reason why

```
test handlers_format::tests::test_format_minimal_diff has been running for over 60 seconds
```

DavisVaughan:
Closes #79 

This is a first pass to completely remove `AUXILIARY_EVENT_TX` in favor of passing around state to access `auxiliary_event_tx`, owned by `GlobalState`.

Currently logging is tied up in auxiliary events, which results in a net loss for ergonomics when you want to log something, as you need to pass `auxiliary_event_tx` around to call its `log_error()` methods.

#93 is a follow up to this PR that splits the log specific events out of this, instead hooking up a tracing subscriber for logging so you can use `tracing::error!()` as a free function anywhere in the LSP

### DIFF
--- a/crates/lsp/src/config.rs
+++ b/crates/lsp/src/config.rs
@@ -95,7 +95,7 @@ impl From<VscDocumentConfig> for DocumentConfig {
                 if var == "tabSize" {
                     x.tab_size
                 } else {
-                    crate::log_warn!("Unknown indent alias {var}, using default");
+                    log::warn!("Unknown indent alias {var}, using default");
                     2
                 }
             }
--- a/crates/lsp/src/handlers_state.rs
+++ b/crates/lsp/src/handlers_state.rs
@@ -33,6 +33,7 @@ use crate::config::DocumentConfig;
 use crate::config::VscDiagnosticsConfig;
 use crate::config::VscDocumentConfig;
 use crate::documents::Document;
+use crate::main_loop::AuxiliaryEventSender;
 use crate::main_loop::LspState;
 use crate::state::workspace_uris;
 use crate::state::WorldState;
@@ -156,6 +157,7 @@ pub(crate) fn did_change(
 pub(crate) fn did_close(
     params: DidCloseTextDocumentParams,
     state: &mut WorldState,
+    auxiliary_event_tx: &AuxiliaryEventSender,
 ) -> anyhow::Result<()> {
     let uri = params.text_document.uri;
 
@@ -167,7 +169,7 @@ pub(crate) fn did_close(
         .remove(&uri)
         .ok_or(anyhow!("Failed to remove document for URI: {uri}"))?;
 
-    crate::log_info!("did_close(): closed document with URI: '{uri}'.");
+    auxiliary_event_tx.log_info(format!("did_close(): closed document with URI: '{uri}'."));
 
     Ok(())
 }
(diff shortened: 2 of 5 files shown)


crates/lsp/src/config.rs
```diff
@@ -101,7 +101,7 @@ impl From<VscDocumentConfig> for DocumentConfig {
                 if var == "tabSize" {
                     x.tab_size
                 } else {
-                    crate::log_warn!("Unknown indent alias {var}, using default");
+                    log::warn!("Unknown indent alias {var}, using default");
```
DavisVaughan, line 104: While its possible to move `auxiliary_events_tx` into here, it didn't feel worth it for this one? Happy to do it though.

crates/lsp/src/handlers_state.rs
```diff
@@ -156,6 +157,7 @@ pub(crate) fn did_change(
 pub(crate) fn did_close(
     params: DidCloseTextDocumentParams,
     state: &mut WorldState,
+    auxiliary_event_tx: &AuxiliaryEventSender,
```
DavisVaughan, line 160: Any LSP request/notification that needs to log should add a `auxiliary_event_tx` argument that is passed through from the top level `handle_event()`

crates/lsp/src/main_loop.rs
```diff
@@ -319,14 +389,51 @@ impl GlobalState {
...
+    }
+
+    /// Spawn a blocking task
+    ///
+    /// This runs tasks that do semantic analysis on a separate thread pool to avoid
+    /// blocking the main loop.
+    ///
+    /// Can optionally return an event for the auxiliary loop (i.e. a log message or
+    /// diagnostics publication).
+    fn spawn_blocking<Handler>(&self, handler: Handler)
```
DavisVaughan: This is no longer a free function but I think that is fine? Our usage of `spawn_blocking()` in ark is all in functions that get called through `GlobalState`, so we can easily pass around `auxiliary_event_tx` to where we need to call `spawn_blocking()`.

It's possible that `spawn_blocking()` should actually be a method on `AuxiliaryEventSender`, as I think that would end up being cleaner and actually make more sense, since this is a special type of sender.

DavisVaughan: Done in https://github.com/posit-dev/air/pull/80/commits/6407db94b2cea63c9e6c44e7596a70941964cf38

lionel- (commented):
My worry is that having to pass state around just for the purpose of logging, to every single helper that might benefit from trace logging, is going to be annoying and cause us not to log in some cases to avoid doing the work or complicating simple looking code.  We've decided offline to keep the global state in regular builds but remove it in tests. We could switch to either log:: or tracing:: and use their global setup facilities to switch between approaches at build time.

DavisVaughan:
Yes, determined we are not doing this. And my preference is to move towards being able to use `log::` or `tracing::` everywhere in air, and having the LSP provide a tracing subscriber that still probably does have a global "sender" variable, which isn't great but seems reasonable and nice from an ergonomic perspective

lionel- (approved):
I wonder what kind of objects we'll send around to spawn tasks but we can figure this out when we get there.

Issue #32, opened 2024-11-14
Handle user requested expansion when the first argument is a hole


DavisVaughan:
I've added this test

```r
# TODO: Holes currently don't force expansion. There is no token attached to
# `RHoleArgument`, so we can't compute the "number of lines before the first token".
fn(
  ,
  x = 1
)
```

Currently formats as

```r
fn(, x = 1)
```

Not sure what the right thing to do here is. We probably do want to force expansion in this weird case? Just for consistency, otherwise there is no way to "request" expansion here. It's a weird one though

This also applies to subset and subset2, where it is more likely to see a leading hole, i.e. `df[, something]`

PR #82, opened 2024-12-05
Implement "leading hole hugging" behavior

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/32

Two main changes:

- Implemented hugging for what I call "leading holes", which greatly improves data.table formatting where its common to have a leading `,` hole for an empty `i` expression (explained below)
- Was forced to sit down and confront comment handling for holes, something I had some TODO comments in there for and knew we'd have to do eventually

# Leading holes

Imagine you have this data.table expression, and you want to expand it over multiple lines

```r
dt2 = dt[, .(total_value_received = sum(price * count)), by = "product_id"]
```

Before this PR, your only option was to hit enter between the `[` and the first `,` and you'd get this:

```r
dt2 = dt[
  , 
  .(total_value_received = sum(price * count)), 
  by = "product_id"
]
```

Not great!

After this PR, the "hole" `[<here>,` is considered a "leading hole". Leading holes "hug" tightly to the opening `[` token no matter what, so you get this:

```r
dt2 = dt[,
  .(total_value_received = sum(price * count)),
  by = "product_id"
]
```

So much nicer! This is pretty common with data.table users, and even chains nicely:

```r
dt2 = ProductReceived[
  Products,
  on = c("product_id" = "id"),
][,
  .(total_value_received = sum(price * count)),
  by = "product_id"
]
```

Note that you can have any number of repeated leading `,`s, they all get aggregated together as a set of leading holes.

If there is a comment within a hole, it is no longer a candidate to be a "leading hole" and is instead considered as a normal argument, subject to typical "fully expanded" rules

```r
dt2 = dt[
	# no longer a leading hole
	,
	.(total_value_received = sum(price * count)),
	by = "product_id"
]
```

To be clear, only holes up front can be leading holes. Trailing and intermediate holes still fully break, like this

```r
dt2 = dt[,
	.(total_value_received = sum(price * count)),
	,
	by = "product_id"
]
```

I think this is good behavior because these holes are relatively "weird" and the extra spacing makes them very visible.

## Leading holes and user requested line breaks

We treat leading holes as roughly invisible when detecting if the user requested a line break. With the example from before:

```r
dt2 = dt[, .(total_value_received = sum(price * count)), by = "product_id"]
```

You can actually add a newline _before or after_ that `,` and it will be treated like a user line break. The `,` will snap back to hugging the `[` tightly but the arguments will get expanded. The principle for user requested expansion here is:

> The user added a line break anywhere before the first non-hole argument in a call

This seems to result in the least surprising behavior.

### DIFF
--- a/crates/air_r_formatter/src/comments.rs
+++ b/crates/air_r_formatter/src/comments.rs
@@ -6,6 +6,7 @@ use air_r_syntax::RIfStatement;
 use air_r_syntax::RLanguage;
 use air_r_syntax::RParenthesizedExpression;
 use air_r_syntax::RSyntaxKind;
+use air_r_syntax::RSyntaxToken;
 use air_r_syntax::RWhileStatement;
 use biome_formatter::comments::CommentKind;
 use biome_formatter::comments::CommentPlacement;
@@ -65,7 +66,8 @@ impl CommentStyle for RCommentStyle {
                 .or_else(handle_if_statement_comment)
                 .or_else(handle_parenthesized_expression_comment)
                 .or_else(handle_argument_name_clause_comment)
-                .or_else(handle_argument_comment),
+                .or_else(handle_argument_comment)
+                .or_else(handle_hole_argument_comment),
             CommentTextPosition::OwnLine => handle_for_comment(comment)
                 .or_else(handle_function_comment)
                 .or_else(handle_while_comment)
@@ -73,7 +75,8 @@ impl CommentStyle for RCommentStyle {
                 .or_else(handle_if_statement_comment)
                 .or_else(handle_parenthesized_expression_comment)
                 .or_else(handle_argument_name_clause_comment)
-                .or_else(handle_argument_comment),
+                .or_else(handle_argument_comment)
+                .or_else(handle_hole_argument_comment),
             CommentTextPosition::SameLine => {
                 // Not applicable for R, we don't have `/* */` comments
                 CommentPlacement::Default(comment)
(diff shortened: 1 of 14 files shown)


crates/air_r_formatter/src/comments.rs
```diff
@@ -351,6 +354,184 @@ fn handle_argument_comment(comment: DecoratedComment<RLanguage>) -> CommentPlace
...
+/// Comment trails `a`. Following token is `,` and previous-sibling of hole is a
+/// non-hole.
+///
+/// ```r
+/// fn(
+///   a,<hole> # comment
+///   ,
+///   b
+/// )
+/// ```
```
DavisVaughan, line 401: Holes and comments are quite tricky to think about. In the following example:

```r
fn(
  a, # comment
  ,
  b
)
```

you might think the formatter assumes the hole is here

```r
fn(
  a, # comment
  <hole>,
  b
)
```

That would make `# comment` trailing on `a` automatically with the `Default` comment placement behavior.

But that isn't what happens! Instead, holes are placed "eagerly" like this:

```r
fn(
  a,<hole># comment
  ,
  b
)
```

This causes `# comment` to be a trailing comment on the hole by default. This trailing comment behavior ALWAYS happens with holes, and is something we exploit as it simplifies our hole comment handling. It is also how the biome group handles array hole comments, so I feel good about it:
https://github.com/biomejs/biome/blob/66458a12a048b0e686b242d9d4dd0f0e90d1f92a/crates/biome_js_formatter/src/comments.rs#L241-L242

PR #86, opened 2024-12-09
Enhance `air format`

DavisVaughan:
Can now perform all of these actions to format a file or set of files:

```shell
air format path/to/directory

air format path/to/file1.R path/to/file2.R

air format path/to/directory1 path/to/directory2
```

Can now perform a `--check` which exits with `1` without writing anything if any files change

```shell
air format path/to/directory --check
```


crates/air/Cargo.toml
```diff
@@ -12,14 +12,22 @@ repository.workspace = true
...
+air_fs = { workspace = true }
 air_r_formatter = { workspace = true }
 air_r_parser = { workspace = true }
 anyhow = { workspace = true }
+biome_console = { workspace = true }
+biome_diagnostics = { workspace = true }
 biome_formatter = { workspace = true }
-clap = { workspace = true }
+biome_parser = { workspace = true }
+clap = { workspace = true, features = ["wrap_help"] }
```
DavisVaughan, line 23: The magic `wrap_help` here enables auto wrapping of help text in the cli
https://github.com/astral-sh/ruff/pull/9633

crates/air/src/args.rs
```diff
@@ -20,15 +20,22 @@ pub(crate) enum Command {
...
-    /// Format a file
+    /// Format a set of files or directories
     Format(FormatCommand),
 }
 
 #[derive(Clone, Debug, Parser)]
 pub(crate) struct LspCommand {}
 
 #[derive(Clone, Debug, Parser)]
+#[command(arg_required_else_help(true))]
```
DavisVaughan, line 31: I really don't like "accidental" formatting since formatting is destructive. It needs to be intentional from the user. So with this `air format` will display the help page for format. Use `air format .` to format the current directory.

I think we should use this for all subcommands tbh

crates/air/src/args.rs
```diff
@@ -20,15 +20,22 @@ pub(crate) enum Command {
...
-    /// The file to format
-    pub file: PathBuf,
+    /// The files or directories to format
+    pub paths: Vec<PathBuf>,
+
+    /// If enabled, format results are not written back to the file. Instead,
+    /// exit with a non-zero status code if any files would have been modified,
+    /// and zero otherwise.
+    #[arg(long)]
+    pub check: bool,
```
DavisVaughan, line 40: `arg(long)` tells clap to allow `--check` but not `-c` for this option (that would be `arg(short)`)

crates/air/src/commands/format.rs
```diff
@@ -1,64 +1,312 @@
...
+    for path in paths {
+        builder.add(path);
+    }
 
-    if parsed.has_errors() {
-        return Ok(ExitStatus::Error);
+    let mut out = Vec::new();
+
+    for path in builder.build() {
+        match path {
```
DavisVaughan, line 119: We utilize BurntSushi's ignore crate (builds on walkdir) to recursively walk the supplied paths (files or directories) and collect all nested R files 

We will eventually add `exclude` and `include` criteria on top of this

We can also eventually make this a parallel walker for performance, but that is harder and can be its own thing

crates/air/src/commands/format.rs
```diff
@@ -1,64 +1,312 @@
...
+                    std::fs::write(&path, new)
+                        .map_err(|err| FormatCommandError::Write(path.clone(), err))?;
+                }
+                FormatMode::Check => {}
+            }
+            Ok(FormatFileResult::Formatted(path))
+        }
+        FormattedSource::Unchanged => Ok(FormatFileResult::Unchanged),
+    }
+}
```
DavisVaughan, line 212: `format_source()` takes `&str` and parses and formats it. It returns `FormattedSource` which nicely tells you if the source is `Unchanged` or not. This is super nice as it lets you optimize by avoiding writing back to file when nothing changed. I think the LSP should also use something like this but I had trouble thinking about exactly what to generalize from here for use in the LSP, since the LSP already has its own copy of the CST that it would prefer to pass in.

crates/air/src/commands/format.rs
```diff
@@ -1,64 +1,312 @@
...
+        })
+        .partition_map(|result| match result {
+            Ok(result) => Either::Left(result),
+            Err(err) => Either::Right(err),
+        });
+
+    for error in &errors {
+        // TODO: Hook up a tracing subscriber!
+        tracing::error!("{error}");
+    }
```
DavisVaughan, line 38: Errors are collected along the way for each path and are emitted at top level here. It's silent right now, but once we hook up a subscriber it will look quite nice as it reports the path and the problem for each error

crates/air/src/commands/format.rs
DavisVaughan: We do have an issue with our string normalization on Windows that prevents this from working well.

- Windows file contains CRLF
- We normalize to LF and store as `source`, but store `LineEnding::CRLF` for the formatter to use
- Formatter emits `formatted` with CRLF endings
- `source` and `formatted` now _always_ look different

I think we should explore not doing the normalization (neither biome nor ruff do it) as I think our tooling can totally handle it, and it would open the door for better optimizations

PR #93, opened 2024-12-11
Use a tracing subscriber for LSP logging

DavisVaughan:
Branched from #80 

The LSP now activates a global tracing "subscriber" for all `tracing::error!()` (and friends) events. This means that we can use the free functions tracing provides to log from anywhere in the LSP (or from the parser/formatter internals) and it will get captured by this global subscriber. (Notably, I have not activated the compatibility helper for `log::`, and have removed that crate).

The global subscriber implements a `write()` method that forwards the message in a non-async way to a new logging specific thread. That logging thread can then call `client.log_message().await` to perform an async send of the message to the client. That's the only job of that thread.

The `AuxiliaryEvent` thread has gotten a bit simpler since we don't worry about logging in there now.

---

Checking for `air.logLevel` and `air.dependencyLogLevels` is not yet done. That is going to be 99% a frontend PR with lots of changes required in the typescript extension, and this PR felt big enough as is. So I'll wait to do that. But here is the general idea:

- User sets `air.logLevel` to `error/info/warn/debug/trace` in their user level `settings.json` (note, ignoring workspace level)
- Forward that value to the server using the `initializationOptions` field of type `LSPAny` within the `InitializeParams`. The client and server will need to have matching representations of what `initializationOptions` looks like, but I'm imagining a `Settings` struct where one field is `logLevel`.
- Server gets that value in the `initialize()` handler and uses it to finalize the `logLevel` and fully set up logging

Note that this approach would work for other IDEs besides VS Code. The equivalent to the air vs code extension would just have to make sure to pass a structured `initializationOptions` field through to the server.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize


crates/lsp/src/handlers_state.rs
```diff
@@ -62,7 +63,13 @@ pub(crate) fn initialize(
     params: InitializeParams,
     lsp_state: &mut LspState,
     state: &mut WorldState,
+    log_tx: LogMessageSender,
 ) -> anyhow::Result<InitializeResult> {
+    // TODO: Get user specified log level from `params.initialization_options`
+    let log_level = None;
+
+    logging::init_logging(log_tx, log_level, params.client_info.as_ref());
```
DavisVaughan, line 71: Initializing logging in the `initialize()` handler is as early as I think we can initialize it. We are going to require information from the Client to correctly set up our logging infrastructure, so we can't do this any earlier.

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// Start the log loop
+    ///
+    /// Takes ownership of log state and start the low-latency log loop.
+    ///
+    /// We use `MessageType::LOG` to prevent the middleware from adding its own
+    /// timestamp and log level labels. We add that ourselves through tracing.
+    pub(crate) async fn start(mut self) {
+        while let Some(message) = self.log_rx.recv().await {
+            self.client
+                .log_message(MessageType::LOG, message.contents)
```
DavisVaughan, line 74: We always forward log messages of type `LOG` to the middleware.

Sending `LOG` messages means the middleware calls the raw `outputChannel.appendLine()` here:
https://github.com/microsoft/vscode-languageserver-node/blob/906f5fb306e1f6059cbdcb1efd962647222b5867/client/src/common/client.ts#L1281-L1297

I prefer that over going through `ERROR` or `WARN` in the middleware, which goes through this `logOutputMessage()` helper:
https://github.com/microsoft/vscode-languageserver-node/blob/906f5fb306e1f6059cbdcb1efd962647222b5867/client/src/common/client.ts#L1179-L1187

That helper unconditionally prepends the log level and a timestamp to the message, but we already do this.

It will also show a notification as a toast message to the user if `showNotification` is `true` or `"force"`, but it's always `false` when processing server log messages. This means that literally the only thing we would get from the middleware by sending `MessageType::ERROR` is for it to automatically prepend `[Error - <time>]` to our messages, so it seems fine to skip this.

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,265 @@
...
+) {
+    let log_level = resolve_log_level(log_level);
+
+    let writer = if client_info.is_some_and(|client_info| {
+        client_info.name.starts_with("Zed") || client_info.name.starts_with("Visual Studio Code")
+    }) {
+        BoxMakeWriter::new(LogWriterMaker::new(log_tx))
+    } else {
+        BoxMakeWriter::new(std::io::stderr)
+    };
```
DavisVaughan, line 144: Doing what ruff does here. It seems like VS Code and Zed handle `window/logMessage` well, but perhaps not every lsp client does? So they specially approve ones that handle it well and send to stderr otherwise.

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,265 @@
...
+        // Display source code line numbers
+        .with_line_number(true)
+        // Don't display the thread ID or thread name
+        .with_thread_ids(false)
+        .with_thread_names(false)
+        // Don't display the event's target (module path).
+        // Mostly redundant with file paths.
+        .with_target(false)
+        // Display local time rather than UTC
+        .with_timer(LocalTime::rfc_3339())
```
DavisVaughan, line 160: Nicely actually displays logs in your local time!

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,265 @@
...
+/// We never log during tests as tests run in parallel within a single process,
+/// but you can only have 1 global subscriber per process.
+///
+/// If you are debugging a single test, you can override this to emit messages to stderr.
+///
+/// Note that if you override this and run multiple tests in parallel, then the call
+/// to `set_global_default()` will error causing a panic.
+fn is_test_client(client_info: Option<&ClientInfo>) -> bool {
+    client_info.map_or(false, |client_info| client_info.name == "AirTestClient")
+}
```
DavisVaughan, line 185: Important notes here about skipping logging during tests. They are very noisy in the test output, and there are issues with running tests in parallel since the "global subscriber" is by nature a "once per process" kind of thing.

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,265 @@
...
+    ) -> bool {
+        // This is a potential reason to use `air_` as the crate prefix,
+        // it would make it easy to set the `tracing_level()` for only air related crates
+        let filter = if meta.target().starts_with("air") || meta.target().starts_with("lsp") {
+            self.filter.tracing_level()
+        } else {
+            tracing::Level::INFO
+        };
+
+        meta.level() <= &filter
```
DavisVaughan, line 259: Remember that tons of other crates use `tracing::info!()` and friends internally

This filter uses `target`, which is typically the module name, to determine the tracing level to filter with, and we never go below `Info` for 3rd party crates.

This is actually one reason it would be quite nice to prefix all of our crates with `air_`! Detection would be very simple.

crates/lsp/src/main_loop.rs
```diff
@@ -288,28 +245,30 @@ impl GlobalState {
...
                 _ => {}
             }
         }
 
-        log::trace!("Main loop closed. Shutting down auxiliary loop.");
+        tracing::trace!("Main loop closed. Shutting down auxiliary and log loop.");
         set.shutdown().await;
-
-        log::trace!("Main loop exited.");
     }
```
DavisVaughan, line 272: Remember that after the log loop is shutdown, we won't be able to see anymore messages. I've removed a few messages that would otherwise occur too soon or too late because of this to prevent confusion.

crates/lsp/src/main_loop.rs
```diff
@@ -360,15 +319,17 @@ impl GlobalState {
...
                     }
                 },
 
                 LspMessage::Request(request, tx) => {
                     match request {
                         LspRequest::Initialize(params) => {
-                            respond(tx, handlers_state::initialize(params, &mut self.lsp_state, &mut self.world), LspResponse::Initialize)?;
+                            // Unwrap: `Initialize` method should only be called once.
+                            let log_tx = self.log_tx.take().unwrap();
+                            respond(tx, handlers_state::initialize(params, &mut self.lsp_state, &mut self.world, log_tx), LspResponse::Initialize)?;
```
DavisVaughan, line 332: I quite like how lsp_server works here. Instead of `Initialize` being an event you handle in your main loop, it is instead handled through explicit calls to `initialize_start()` to receive the `Initialize` request, and `initialize_finish()` to send back an `InitializeResult`. 

I like that because it really drives home that they are once-per-session methods, not ones that can be recalled through the main loop.

https://docs.rs/lsp-server/latest/lsp_server/struct.Connection.html#method.initialize_start

crates/lsp/src/tower_lsp.rs
```diff
@@ -188,27 +179,20 @@ where
     I: AsyncRead + Unpin,
     O: AsyncWrite,
 {
-    log::trace!("Starting LSP");
-
     let (service, socket) = new_lsp();
     let server = tower_lsp::Server::new(read, write, socket);
     server.serve(service).await;
-
-    log::trace!("LSP exiting gracefully.",);
```
DavisVaughan, line 197: Again these messages are "too soon" or "too late" from the perspective of the log thread

crates/lsp_test/src/lsp_client.rs
```diff
@@ -87,9 +88,22 @@ impl TestClient {
...
+    // `AirTestClient` so we can recognize it when we set up global logging.
+    fn with_client_info(
+        mut init_params: lsp_types::InitializeParams,
+    ) -> lsp_types::InitializeParams {
+        init_params.client_info = Some(ClientInfo {
+            name: String::from("AirTestClient"),
+            version: None,
+        });
+        init_params
+    }
```
DavisVaughan, line 105: I like using `AirTestClient` as the way to determine "are we testing or not" rather than `cfg(debug_assertions)` because this will also "look like" testing mode in integration tests!

crates/lsp/src/logging.rs
lionel-: That seems strange to me. It feels like it would be better to control this with the log level. By default we don't see this output since it's captured anyway.

Tests are run in parallel that's true, but I don't think it's customary to disable logging in tests altogether? It seems like we're making it harder to debug tests here.

crates/lsp/src/logging.rs
lionel-: I think the filter implemented in ark is more flexible? https://github.com/posit-dev/ark/blob/3b8ff6112d14f1b78640e4db55c82754c0b362bf/crates/ark/src/logger.rs#L26-L43

Is there any reason not to use this approach?

crates/lsp/src/tower_lsp.rs
```diff
@@ -102,12 +97,12 @@ impl Backend {
         // Wait for response from main loop
         let out = response_rx.recv().await.unwrap()?;
 
-        self.log_info(format!("Outgoing {out:#?}"));
+        tracing::trace!("Outgoing\n{out:#?}");
```
lionel-, line 100:
```suggestion
        tracing::trace!("Outgoing:\n{out:#?}");
```

crates/lsp/src/tower_lsp.rs
lionel-: Could be `eprintln!()` then

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,324 @@
...
+//!
+//! Resolution strategy:
+//!
+//! - The environment variable `AIR_DEPENDENCY_LOG_LEVELS` is consulted.
+//!
+//! - The LSP `InitializeParams.initializationOptions.dependencyLogLevels` option is
+//!   consulted. This can be set in VS Code / Positron using `air.dependencyLogLevel`, or
+//!   in Zed by supplying `initialization_options`.
+//!
+//! - If neither are supplied, we fallback to no logging for dependency crates.
```
DavisVaughan, line 37: Here is the newly updated resolution strategy

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,324 @@
...
+// TODO: Is there a way to generate this at compile time?
+const TARGETS: &[&str] = &[
+    "air_r_factory",
+    "air_r_formatter",
+    "air_r_parser",
+    "air_r_syntax",
+    "fs",
+    "line_ending",
+    "lsp",
+];
```
DavisVaughan, line 235: Something rather interesting is that even though we target these crate names, really this ends up being a _prefix match_ on the log `target` field. That means our `lsp_test` crate matches and we got lots of logs around `codec.rs`, so I just commented those out for now.

---

We could actually take advantage of the prefix matching if we wanted to. By prefixing all of our crates with `air_`, this would simplify the code down to a single:

```
filter.with_target("air_", log_level)
```

Which I like because:
- We don't have to worry about updating this list
- We worry less about prefix matching a crate that isn't us

DavisVaughan (commented):
Okay after some reworking and talking to @lionel- here is where we landed:

- `cargo test` is silent, no logs. Aliased to `just test`.

- `cargo test -- --nocapture` shows logs, but you really want to run with `--test-threads 1` so the logging is scoped to the test it applies to. Use `just test-verbose` for this.

- `AIR_LOG_LEVEL=trace` allows you to set the log level of air crates. `air.logLevel` will be the VS Code option. The option will be an enum with autocomplete.

- `AIR_DEPENDENCY_LOG_LEVELS=tokio=info,tower_lsp=debug` allows you to set the log level of 3rd party crates. `air.dependencyLogLevels` will be the VS Code option. The option will be free text.

- Note that by default, we do `Info` logs for air crates and NO logging for 3rd party crates.

- CI runs with `cargo test -- --nocapture --test-threads 1` and `AIR_LOG_LEVEL=trace`. This should make it pretty helpful if something goes wrong on CI.

crates/lsp/src/tower_lsp.rs
DavisVaughan: I feel like it would be quite odd to have two messages get sent to stderr when all other messages end up over `window/logMessage`. I feel like it just adds one more layer of places to look for logs. Would it be ok to try and drop these for now and see if we need them later on?

crates/lsp/src/tower_lsp.rs
```diff
@@ -65,31 +66,97 @@ pub(crate) enum LspRequest {
...
+impl std::fmt::Display for LspRequest {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.into())
+    }
+}
+impl std::fmt::Display for LspResponse {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.into())
+    }
+}
```
DavisVaughan, line 91: Using `strum::IntoStaticStr`, we make the `Display` method show just the enum variant name

crates/lsp/src/tower_lsp.rs
```diff
@@ -65,31 +66,97 @@ pub(crate) enum LspRequest {
...
+                f.debug_tuple(self.inner.into())
+                    .field(&params.text_document.uri)
+                    .field(&params.text_document.version)
+                    .field(&params.text_document.language_id)
+                    .finish()
+            }
+            _ => std::fmt::Debug::fmt(self.inner, f),
+        }
+    }
+}
```
DavisVaughan, line 133: It ended up being easier to write a wrapper type that overrides specific variants but otherwise pushes through to the "normal" derived `Debug` implementation. If we try and implement `Debug` on `LspNotification` directly then we lose everything it auto-derives for us, which is actually quite a lot of repetitive code.

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,324 @@
...
+use tracing_subscriber::fmt::time::LocalTime;
+use tracing_subscriber::fmt::TestWriter;
+use tracing_subscriber::{
+    fmt::{writer::BoxMakeWriter, MakeWriter},
+    layer::SubscriberExt,
+    Layer,
+};
+
+// TODO:
+// - Add `air.logLevel` and `air.dependencyLogLevels` as VS Code extension options that set
```
DavisVaughan, line 63: I'm also considering going the rust-analyzer route of having lots of smaller namespaces for options, like in this case:

```
air.log.level
air.log.dependencyLevels
```

which would come in as a `LogSettings` struct containing `level` and `dependencyLevels` fields.

That seems like it would be useful for grouping related options?

crates/lsp/src/tower_lsp.rs
lionel-: cool!

crates/lsp/src/tower_lsp.rs
```diff
@@ -65,31 +66,97 @@ pub(crate) enum LspRequest {
...
+    inner: &'a LspNotification,
+}
+struct TraceLspRequest<'a> {
+    inner: &'a LspRequest,
+}
+struct TraceLspResponse<'a> {
+    inner: &'a LspResponse,
+}
+
+impl std::fmt::Debug for TraceLspNotification<'_> {
```
lionel-, line 119: 👍

crates/lsp/src/tower_lsp.rs
lionel-: I was thinking about this problem after our call yesterday. That's a nice solution.

crates/lsp/src/tower_lsp.rs
lionel-: Sure that's fine with me.

justfile
```diff
@@ -6,13 +6,18 @@ gen-formatter:
...
 test:
-  cargo test -p air_r_parser
-  cargo test -p air_r_formatter
+  cargo test
+
+# Run the tests in verbose mode
+# `--nocapture` to see our own `tracing` logs
+# `--test-threads 1` to ensure `tracing` logs aren't interleaved
+test-verbose:
+  cargo test -- --nocapture --test-threads 1
```
lionel-, line 17: I like this

crates/lsp/src/tower_lsp.rs
```diff
@@ -65,31 +66,97 @@ pub(crate) enum LspRequest {
...
-
     /// Handle to main loop. Drop it to cancel the loop, all associated tasks,
     /// and drop all owned state.
     _main_loop: tokio::task::JoinSet<()>,
 }
 
 impl Backend {
     async fn request(&self, request: LspRequest) -> anyhow::Result<LspResponse> {
-        self.log_info(format!("Incoming: {request:#?}"));
+        tracing::trace!("Incoming:\n{request:#?}", request = request.trace());
```
lionel-, line 159: I think we also discussed a `tracing::info()` message with the message type so we can better make sense of user logs before they enable trace logging?

crates/lsp/src/logging.rs
```diff
@@ -0,0 +1,324 @@
...
+/// Important notes:
+/// - `cargo test` swallows all logs unless you use `-- --nocapture`.
+/// - Tests run in parallel, so logs can be interleaved unless you run `--test-threads 1`.
+///
+/// We use `cargo test -- --nocapture --test-threads 1` on CI because of all of this.
+fn is_test_client(client_info: Option<&ClientInfo>) -> bool {
+    client_info.map_or(false, |client_info| client_info.name == "AirTestClient")
+}
+
+// TODO: Is there a way to generate this at compile time?
```
lionel-, line 226: chatgpt suggests

```sh
cargo metadata --no-deps --format-version 1 | jq -r '.packages[].name'
```

which gives me

```
air
air_r_formatter
air_r_syntax
air_formatter_test
line_ending
air_r_parser
air_r_factory
tests_macros
fs
lsp
lsp_test
biome_ungrammar
xtask_codegen
xtask
```

I guess we can add an exclude list for build-time things like biome_ungrammar. Or there's probably a way to determine dev deps from the metadata.

lionel-: If you remove the jq you get a bunch of json data which is fine to ingest from Rust. This could be done in `build.rs`.

.github/workflows/test-linux.yml
```diff
@@ -51,4 +51,8 @@ jobs:
         run: cargo build
 
       - name: Run unit tests
-        run: cargo test
+        env:
+          AIR_LOG_LEVEL: trace
+        # `--nocapture` to see our own `tracing` logs
+        # `--test-threads 1` to ensure `tracing` logs aren't interleaved
+        run: cargo test -- --nocapture --test-threads 1
```
lionel-, line 58: Use the `just` alias? I don't mind either way, I guess this is more explicit.

lionel- (approved):
Looks great!

My only comment is that I think we should unconditionally log the LSP message types at info level.

crates/lsp/src/logging.rs
DavisVaughan: Of course there's a crate for this. `cargo_metadata`! https://github.com/posit-dev/air/pull/93/commits/3f0f72058866a26f6316dd2ad26939fea769ee8e

PR #109, opened 2024-12-16
Read `air.toml` and use in `air format`

### LINKED ISSUE #100: Hook up user specified options of `air.logLevel` and `air.dependencyLogLevels` in VS Code extension
The server side is mostly ready to go, but we need to create client and server representations of "client settings" like

```
ClientSettings {
  logLevel: Option<LogLevel>,
  dependencyLogLevels: Option<String>
}
```

Note that every field in `ClientSettings` should be optional. This allows:
- Easier backwards compat
- Partial initialization like in Zed's `initialization_options` setting

We will pass the settings through `InitializeParams.initializationOptions`, which is an "any" type field we are allowed to control

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/83
Closes #100 

TODO: Need to update this description with new changes and possibly split into 2 prs (the 2 commits represent the 2 prs)

---

A first pass of support for `air.toml`

At a high level:

- `air.toml` supported file structure is outlined by the `TomlOptions` struct. _Every_ field in this struct is `Option<>` guarded.

- `TomlOptions::into_settings()` converts a `TomlOptions` into a finalized `Settings`, using defaults when nothing was set. `Settings` is more friendly for internal use than for user facing structure.

- With `air format dir1 dir2`, for both `dir1` and `dir2` we find the "closest" `air.toml` for each path and save the directory it was found in as, say, `ancestor1` and `ancestor2`. We read in the `air.toml` and associate the `Settings` with that ancestor directory using a `Resolver`.

- We iterate through `dir1` and `dir2` collecting a flat vector of R files. Then we iterate over each R file path, retrieve its associated `Settings` from the `Resolver`, and format with that path + settings combination.

### DIFF
--- a/crates/workspace/src/resolve.rs
+++ b/crates/workspace/src/resolve.rs
@@ -1,0 +1,252 @@
+// --- source
+// authors = ["Charlie Marsh"]
+// license = "MIT"
+// origin = "https://github.com/astral-sh/ruff/tree/main/crates/ruff_workspace"
+// ---
+
+use std::collections::btree_map::Keys;
+use std::collections::btree_map::Range;
+use std::collections::btree_map::RangeMut;
+use std::collections::BTreeMap;
+use std::path::{Path, PathBuf};
+
+use ignore::DirEntry;
+use rustc_hash::FxHashSet;
+use thiserror::Error;
+
+use crate::settings::Settings;
+use crate::toml::find_air_toml_in_directory;
+use crate::toml::parse_air_toml;
+use crate::toml::ParseTomlError;
+
+/// Resolves a [`Path`] to its associated `T`
+///
+/// To use a [`PathResolver`]:
+/// - Load directories into it using [`PathResolver::add()`]
+/// - Resolve a [`Path`] to its associated `T` with [`PathResolver::resolve()`]
+///
+/// See [`PathResolver::resolve()`] for more details on the implementation.
+#[derive(Debug, Default)]
+pub struct PathResolver<T> {
+    /// Fallback value to be used when a `path` isn't associated with anything in the `map`
+    fallback: T,
+
+    /// An ordered `BTreeMap` from a `path` (normally, a directory) to a `T`
+    map: BTreeMap<PathBuf, T>,
+}
+
+impl<T> PathResolver<T> {
+    /// Create a new empty [`PathResolver`]
+    pub fn new(fallback: T) -> Self {
+        Self {
+            fallback,
+            map: BTreeMap::new(),
+        }
+    }
+
+    pub fn fallback(&self) -> &T {
+        &self.fallback
+    }
+
+    pub fn add(&mut self, path: &Path, value: T) -> Option<T> {
+        self.map.insert(path.to_path_buf(), value)
+    }
+
+    pub fn remove(&mut self, path: &Path) -> Option<T> {
+        self.map.remove(path)
+    }
+
+    pub fn len(&self) -> usize {
+        self.map.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.map.is_empty()
+    }
+
+    pub fn keys(&self) -> Keys<'_, PathBuf, T> {
+        self.map.keys()
+    }
+
+    pub fn clear(&mut self) {
+        self.map.clear();
+    }
+
+    /// Resolve a [`Path`] to its associated `T`
+    ///
+    /// This resolver works by finding the closest directory to the `path` to search for.
+    ///
+    /// The [`BTreeMap`] is an ordered map, so if you do:
+    ///
+    /// ```text
+    /// resolver.add("a/b", value1)
+    /// resolver.add("a/b/c", value2)
+    /// resolver.add("a/b/d", value3)
+    /// resolver.resolve("a/b/c/test.R")
+    /// ```
+    ///
+    /// Then it detects both `"a/b"` and `"a/b/c"` as being "less than" the path of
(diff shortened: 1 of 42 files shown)


crates/workspace/src/resolve.rs
```diff
@@ -0,0 +1,192 @@
...
+    // to remind us to expose them.
+    //
+    // "This toggles, as a group, all the filters that are enabled by default"
+    // builder.standard_filters(true)
+    builder.hidden(true);
+    builder.parents(true);
+    builder.ignore(true);
+    builder.git_ignore(true);
+    builder.git_global(true);
+    builder.git_exclude(true);
```
DavisVaughan, line 149: It's nice that we will get gitignore support out of the box so we don't format any directories marked in a .gitignore

crates/workspace/src/settings.rs
```diff
@@ -0,0 +1,60 @@
+mod indent_style;
+mod indent_width;
+// TODO: Can we pick a better crate name for `line_ending` so these don't collide?
+#[path = "settings/line_ending.rs"]
+mod line_ending_setting;
```
DavisVaughan, line 5: Mildly annoying. We have a `line_ending` crate but I've also added `settings/line_ending.rs` as a mod and I need to use both in here 😢 .

crates/workspace/src/settings.rs
```diff
@@ -0,0 +1,60 @@
...
+        };
+
+        RFormatOptions::new()
+            .with_indent_style(self.indent_style.into())
+            .with_indent_width(self.indent_width.into())
+            .with_line_ending(line_ending)
+            .with_line_width(self.line_length.into())
+            .with_magic_line_break(self.magic_line_break.into())
+    }
+}
```
DavisVaughan, line 60: Finalized `FormatSettings` are really "mostly" finalized. Things like `LineEnding::Auto` get truly finalized when converting to `RFormatOptions`, because that's when we are most likely to have the source code to infer line endings for lying around.

crates/workspace/src/settings/line_ending.rs
```diff
@@ -0,0 +1,31 @@
+use std::fmt;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Default, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum LineEnding {
```
DavisVaughan, line 5: I think it is very reasonable / inevitable that we end up with our own copies of these settings in addition to the ones from biome.

- For each one we need a `Deserialize` method that works nicely with the toml crate and uses kebab-case
- For some, like here, we add additional user facing options like `Auto` that don't make it into the underlying formatter option
- For some, like `IndentWidth`, we may choose to have our own `IndentWidth::default()` that is different from what biome's is

crates/workspace/src/toml.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::path::{Path, PathBuf};
+
+/// Parse an `air.toml` file.
+pub fn parse_air_toml<P: AsRef<Path>>(path: P) -> Result<TomlOptions, ParseTomlError> {
+    let contents = std::fs::read_to_string(path.as_ref())
+        .map_err(|err| ParseTomlError::Read(path.as_ref().to_path_buf(), err))?;
+
+    toml::from_str(&contents)
+        .map_err(|err| ParseTomlError::Deserialize(path.as_ref().to_path_buf(), err))
+}
```
DavisVaughan, line 16: This is the entire set of parsing code. It's wild how little it is. Everything is set up at compile time instead!

crates/workspace/src/toml_options.rs
```diff
@@ -0,0 +1,121 @@
...
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct GlobalTomlOptions {
+    /// The line length at which the formatter prefers to wrap lines.
+    ///
+    /// The value must be greater than or equal to `1` and less than or equal to `320`.
+    ///
+    /// Note: While the formatter will attempt to format lines such that they remain
+    /// within the `line-length`, it isn't a hard upper bound, and formatted lines may
+    /// exceed the `line-length`.
+    pub line_length: Option<LineLength>,
```
DavisVaughan, line 40: ruff has some very cool macros that use `TomlOptions` as its "source of truth" for documentation.

It extracts out each table name (like `format`) and each option under the table (like `line_length`) along with all documentation comments and makes a really nice markdown page for it that ends up here:

https://docs.astral.sh/ruff/settings/

Without a page like this its hard to know what the spec is for what you can supply in `air.toml`

crates/workspace/src/toml_options.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// The number of spaces per indentation level (tab).
+    ///
+    /// The value must be greater than or equal to `1` and less than or equal to `24`.
+    ///
+    /// Used by the formatter to determine the visual width of a tab.
+    ///
+    /// This option changes the number of spaces the formatter inserts when
+    /// using `indent-style = "space"`. It also represents the width of a tab when
+    /// `indent-style = "tab"` for the purposes of computing the `line-length`.
+    pub indent_width: Option<IndentWidth>,
```
DavisVaughan, line 51: Both `line_length` and `indent_width` are top-level global options used by "all of air" because if we consider a linter then it will need access to these options to determine if a line exceeds the line length.

Compare that with `indent_style`, `line_ending`, and `ignore_magic_line_break` which are more clearly format specific and live under the `[format]` table header.

PR #121, opened 2024-12-20
Send structured `InitializationOptions` to the server

DavisVaughan:
In particular, this lets us respect `air.logLevel` and `air.dependencyLogLevels` on server startup

User and workspace level settings are not used yet but are included for completeness

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -6,6 +6,10 @@
 	"publisher": "Posit",
 	"license": "MIT",
 	"repository": "https://github.com/posit-dev/air",
+	"serverInfo": {
+		"name": "Air",
+		"module": "air"
+	},
 	"engines": {
 		"vscode": "^1.90.0"
 	},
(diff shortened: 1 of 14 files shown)


editors/code/package.json
```diff
@@ -35,6 +35,29 @@
 				]
 			}
 		],
+		"configuration": {
+			"properties": {
+				"air.logLevel": {
+					"default": null,
+					"markdownDescription": "Controls the log level of the language server.\n\n**This setting requires a restart to take effect.**",
```
lionel-, line 42: These static settings are a bit annoying. Probably worth it to make it dynamic right away?

editors/code/src/settings.ts
```diff
@@ -0,0 +1,41 @@
+import { ConfigurationScope, workspace, WorkspaceConfiguration } from "vscode";
+
+type LogLevel = "error" | "warn" | "info" | "debug" | "trace";
+
+// This is a direct representation of the Client settings sent to the Server in the
+// `initializationOptions` field of `InitializeParams`. These are only pulled at the
+// user level since they are global settings on the server side (and are scoped to
+// `"scope": "application"` in `package.json` so they can't even be set at workspace level).
+export type IInitializationOptions = {
```
lionel-, line 9:
```suggestion
export type InitializationOptions = {
```
?

editors/code/src/settings.ts
```diff
@@ -0,0 +1,41 @@
...
+	logLevel?: LogLevel;
+	dependencyLogLevels?: string;
+};
+
+export async function getInitializationOptions(
```
lionel-, line 14: Does this need to be async?

lionel- (commented):
Did you know about `handle_initialized()` in handlers.rs? We already have some infra in place for dynamic watching of settings. Probably worth making these dynamic now?

PR #122, opened 2024-12-20
Add support for reading `air.toml` in the CLI and LSP

### LINKED ISSUE #83: Add a basic configuration parser for `air.toml`

DavisVaughan:
Closes #83

The point of this PR is to allow air to read `air.toml` configuration files in your project and pass information from them on to format document / selection requests. There are 3 main parts:

- Finding and reading `air.toml` files. This is all contained in a new `workspace` crate. We read `air.toml` files with the `toml` crate. The "spec" is the `TomlOptions` struct itself.

- Utilizing the workspace tools in the CLI. This is pretty easy and small since each CLI call is a "one shot" discovery of `air.toml` configuration files (i.e. there is no state to maintain between CLI calls).

- Utilizing the workspace tools in the LSP. This is much more involved because we want to:
    - Parse and cache the `air.toml`s we find and associate them correctly with R files in a workspace
    - Refresh the `air.toml`s when the user adds a new workspace, or if the user changes the `air.toml`


crates/lsp/src/capabilities.rs
```diff
@@ -0,0 +1,40 @@
...
+use tower_lsp::lsp_types::PositionEncodingKind;
+
+/// A resolved representation of the [ClientCapabilities] the Client sends over that we
+/// actually do something with
+#[derive(Debug, Default)]
+pub(crate) struct ResolvedClientCapabilities {
+    pub(crate) position_encodings: Vec<PositionEncodingKind>,
+    pub(crate) dynamic_registration_for_did_change_configuration: bool,
+    pub(crate) dynamic_registration_for_did_change_watched_files: bool,
+}
```
DavisVaughan, line 11: Collecting client capabilities we care about here. We get them in `Initialize` but use them in other places

crates/lsp/src/handlers.rs
```diff
@@ -24,30 +26,55 @@ pub(crate) async fn handle_initialized(
...
+                    watchers: vec![FileSystemWatcher {
+                        glob_pattern: lsp_types::GlobPattern::String("**/air.toml".into()),
+                        kind: None,
+                    }],
+                })
+                .unwrap(),
+            ),
+        };
+
+        registrations.push(watch_air_toml_registration);
```
DavisVaughan, line 73: Watching for changes in `air.toml` files anywhere within a workspace (creation, deletion, modify)

crates/lsp/src/handlers_format.rs
```diff
@@ -5,35 +5,32 @@
...
     let doc = state.get_document(&params.text_document.uri)?;
 
-    let line_width = LineWidth::try_from(80).map_err(|err| anyhow::anyhow!("{err}"))?;
-
-    // TODO: Handle FormattingOptions
-    let options = RFormatOptions::default()
-        .with_indent_style(IndentStyle::Space)
-        .with_line_width(line_width);
+    let settings = lsp_state.document_settings(&params.text_document.uri);
+    let format_options = settings.format.to_format_options(&doc.contents);
```
DavisVaughan, line 27: After all was said and done this is the main change to the `FormatDocument` request code, pretty nice right now

Note that we are currently ignoring anything related to the document settings provided by the editor

crates/lsp/src/handlers_state.rs
```diff
@@ -199,6 +191,36 @@ pub(crate) async fn did_change_configuration(
...
+
+    for lsp_types::WorkspaceFolder { uri, .. } in params.event.added {
+        errors.push_err(lsp_state.open_workspace_folder(&uri, Settings::default()));
+    }
+    for lsp_types::WorkspaceFolder { uri, .. } in params.event.removed {
+        errors.push_err(lsp_state.close_workspace_folder(&uri));
+    }
+
+    errors.into_result()
+}
```
DavisVaughan, line 209: Updating our per-workspace settings resolver when a workspace is opened / closed

crates/lsp/src/handlers_state.rs
```diff
@@ -199,6 +191,36 @@ pub(crate) async fn did_change_configuration(
...
+    lsp_state: &mut LspState,
+) -> anyhow::Result<()> {
+    for change in &params.changes {
+        lsp_state
+            .workspace_settings_resolver
+            .reload_workspaces_matched_by_url(&change.uri);
+    }
+
+    Ok(())
+}
```
DavisVaughan, line 222: Updating our per-workspace settings resolver when an `air.toml` is changed

crates/lsp/src/snapshots/lsp__handlers_format__tests__format_range_logical_lines-4.snap
```diff
@@ -4,5 +4,5 @@ expression: output
 ---
 1+1
 {
-  2 + 2
+	2 + 2
```
DavisVaughan, line 7: The LSP was hardcoded to `Space` before but we are going to get the default settings now, which are currently `Tab`

crates/lsp/src/workspaces.rs
```diff
@@ -0,0 +1,174 @@
...
+        if url.scheme() == "untitled" && self.path_to_settings_resolver.len() == 1 {
+            tracing::trace!("Using workspace settings for 'untitled' URL: {url}");
+            let workspace_path = self.path_to_settings_resolver.keys().next().unwrap();
+            let path = workspace_path.join("untitled");
+            return self.settings_for_path(&path);
+        }
+
+        tracing::trace!("Using default settings for non-file URL: {url}");
+        self.path_to_settings_resolver.fallback().fallback()
+    }
```
DavisVaughan, line 109: Tricky little behavior here with untitled files

If there aren't any workspaces open, then the untitled file can't be associated with any `air.toml`

If there are >1 workspaces open, we can't make a reasonable guess about which workspace's `air.toml` to try and associate the untitled file with

If there is exactly 1 workspace open (the 95% case), then we can make a reasonable guess that we should use that workspace's `air.toml` with untitled files

crates/lsp/src/workspaces.rs
```diff
@@ -0,0 +1,174 @@
...
+            settings_resolver.clear();
+
+            if let Err(error) = settings_resolver.load_from_paths(&[workspace_path]) {
+                tracing::error!(
+                    "Failed to reload workspace settings for {path}:\n{error}",
+                    path = workspace_path.display(),
+                    error = error
+                );
+            }
+        }
```
DavisVaughan, line 148: I find this "refresh" behavior kind of awesome. We mutably iterate over the workspace specific resolvers that are associated with the updated `air.toml` and _update them in place_ by clearing them and then rediscovering the `air.toml`s within them

crates/workspace/src/resolve.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// resolver.add("a/b/d", value3)
+    /// resolver.resolve("a/b/c/test.R")
+    /// ```
+    ///
+    /// Then it detects both `"a/b"` and `"a/b/c"` as being "less than" the path of
+    /// `"a/b/c/test.R"`, and then chooses `"a/b/c"` because it is at the back of
+    /// that returned sorted list (i.e. the "closest" match).
+    pub fn resolve(&self, path: &Path) -> Option<&T> {
+        self.resolve_entry(path).map(|(_, value)| value)
+    }
```
DavisVaughan, line 93: This is the main thing to read about how `path/to/file.R -> Settings` resolution works

crates/workspace/src/resolve.rs
```diff
@@ -0,0 +1,252 @@
...
+    // to remind us to expose them.
+    //
+    // "This toggles, as a group, all the filters that are enabled by default"
+    // builder.standard_filters(true)
+    builder.hidden(true);
+    builder.parents(true);
+    builder.ignore(false);
+    builder.git_ignore(true);
+    builder.git_global(true);
+    builder.git_exclude(true);
```
DavisVaughan, line 209: It's nice that we will get gitignore support out of the box so we don't format any directories marked in a .gitignore

crates/workspace/src/settings.rs
```diff
@@ -0,0 +1,60 @@
...
+            },
+        };
+
+        RFormatOptions::new()
+            .with_indent_style(self.indent_style.into())
+            .with_indent_width(self.indent_width.into())
+            .with_line_ending(line_ending)
+            .with_line_width(self.line_length.into())
+            .with_magic_line_break(self.magic_line_break.into())
+    }
```
DavisVaughan, line 59: Finalized `FormatSettings` are really "mostly" finalized. Things like `LineEnding::Auto` get truly finalized when converting to `RFormatOptions`, because that's when we are most likely to have the source code to infer line endings for lying around.

crates/workspace/src/settings/line_ending.rs
```diff
@@ -0,0 +1,31 @@
+use std::fmt;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Default, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum LineEnding {
```
DavisVaughan, line 5: I think it is very reasonable / inevitable that we end up with our own copies of these settings in addition to the ones from biome.

- For each one we need a `Deserialize` method that works nicely with the toml crate and uses kebab-case
- For some, like here, we add additional user facing options like `Auto` that don't make it into the underlying formatter option
- For some, like `IndentWidth`, we may choose to have our own `IndentWidth::default()` that is different from what biome's is

crates/workspace/src/toml_options.rs
```diff
@@ -0,0 +1,121 @@
...
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct GlobalTomlOptions {
+    /// The line length at which the formatter prefers to wrap lines.
+    ///
+    /// The value must be greater than or equal to `1` and less than or equal to `320`.
+    ///
+    /// Note: While the formatter will attempt to format lines such that they remain
+    /// within the `line-length`, it isn't a hard upper bound, and formatted lines may
+    /// exceed the `line-length`.
+    pub line_length: Option<LineLength>,
```
DavisVaughan, line 40: ruff has some very cool macros that use `TomlOptions` as its "source of truth" for documentation.

It extracts out each table name (like `format`) and each option under the table (like `line_length`) along with all documentation comments and makes a really nice markdown page for it that ends up here:

https://docs.astral.sh/ruff/settings/

Without a page like this its hard to know what the spec is for what you can supply in `air.toml`

crates/workspace/src/toml_options.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// The number of spaces per indentation level (tab).
+    ///
+    /// The value must be greater than or equal to `1` and less than or equal to `24`.
+    ///
+    /// Used by the formatter to determine the visual width of a tab.
+    ///
+    /// This option changes the number of spaces the formatter inserts when
+    /// using `indent-style = "space"`. It also represents the width of a tab when
+    /// `indent-style = "tab"` for the purposes of computing the `line-length`.
+    pub indent_width: Option<IndentWidth>,
```
DavisVaughan, line 51: Both `line_length` and `indent_width` are top-level global options used by "all of air" because if we consider a linter then it will need access to these options to determine if a line exceeds the line length.

Compare that with `indent_style`, `line_ending`, and `ignore_magic_line_break` which are more clearly format specific and live under the `[format]` table header.

editors/code/src/lsp.ts
```diff
@@ -65,11 +65,6 @@ export class Lsp {
 				{ language: "r", pattern: "**/*.{r,R}" },
 				{ language: "r", pattern: "**/*.{rprofile,Rprofile}" },
 			],
-			synchronize: {
-				// Notify the server about file changes to R files contained in the workspace
-				fileEvents:
-					vscode.workspace.createFileSystemWatcher("**/*.[Rr]"),
-			},
```
DavisVaughan, line 72: This causes the middleware to automatically start sending `DidChangeWatchedFiles` notifications to the Client

It was super confusing for me when I was registering the `air.toml` watched files notification because it looked like I had somehow done it wrong and was getting `.R` files too!

crates/lsp/src/error.rs
```diff
@@ -0,0 +1,52 @@
+/// A tool for collecting multiple anyhow errors into a single [`anyhow::Result`]
```
lionel-, line 1: Looks useful

crates/lsp/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!(self))
+        }
+    }
+}
+
+impl std::error::Error for ErrorVec {}
+
+impl std::fmt::Display for ErrorVec {
```
lionel-, line 40: Probably also needs `Debug`?

crates/lsp/src/handlers.rs
lionel-: This will eventually be replaced by a VFS à la rust-analyzer right?

crates/lsp/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /// Push a new error to the error vector
+    pub(crate) fn push(&mut self, error: anyhow::Error) {
+        self.errors.push(error);
+    }
+
+    /// Convert a error vector into a single [`anyhow::Result`] that knows how to print
+    /// each of the individual errors
+    pub(crate) fn into_result(self) -> anyhow::Result<()> {
```
lionel-, line 29: Probably worth taking an argument and making generic in the `Ok` type?

crates/workspace/src/toml_options.rs
lionel-: We'll also need `indent_style` for format-as-you-type but I guess it makes sense to have it in that section too.

Refactors might need it as well, but I guess we'll reformat refactored pieces, so that will go through the formatter?

lionel-: From a user perspective it seems inconsistent to have format options in different sections though. Is there no way to access format options from diagnostics?

editors/code/src/lsp.ts
lionel-: If we add diagnostics without VFS (as currently implemented in Ark) we'll need to re-enable this.

crates/workspace/src/toml_options.rs
lionel-: Nice

lionel- (approved):
I only did a high level review. LGTM!

crates/lsp/src/handlers.rs
DavisVaughan: AFAICT rust-analyzer still uses this as a signal that the Cargo.toml has changed and needs to be reloaded
https://github.com/rust-lang/rust-analyzer/blob/91fc0a239af4e56b84b1d3974ac0f34dcc99b895/crates/rust-analyzer/src/reload.rs#L538

https://github.com/rust-lang/rust-analyzer/blob/91fc0a239af4e56b84b1d3974ac0f34dcc99b895/crates/rust-analyzer/src/handlers/notification.rs#L276-L286

crates/workspace/src/toml_options.rs
DavisVaughan: I guess there's a question of whether or not our `air.toml` should have sub headers at all 🤔

editors/code/src/lsp.ts
DavisVaughan: Can you help me understand more about why this would be required? Why is getting `didChange` on `.R` files not enough?

DavisVaughan:
Main TODO before merging is deciding whether or not to put everything under `[format]` for now

editors/code/src/lsp.ts
lionel-: It may be, the main thing is to make sure there are no race conditions. I think Rust-analyzer tracks all non-open files, including workspace ones, via the VFS. Possibly for consistency reasons, possibly because that's just as convenient, and possibly for performance reasons (OS notifications rather than LSP ones).

lionel-:
By the way some new files are missing a (c) header

PR #126, opened 2025-01-03
Rewrite the language server on top of a `ruff_server` fork

DavisVaughan:
This PR is a huge move away from tower-lsp, instead implementing our language server on top of lsp-types and lsp-server directly. I believe this has some really nice advantages, and is something we have been discussing for some time now.

In particular, I've implemented this by effectively forking `ruff_server`, dropping everything we don't need, and layering in all of our own additions (`viewFile`, test clients, minimal text edits in format responses, etc).

Some huge benefits:

- Drops all async code from the language server. We now use a scheduler that handles requests/notifications in one of 3 ways:
    - Sync requests/notifs are handled on the main thread, and block
    - Background requests/notifs are handled on a background thread with its own thread pool (typically 4 threads)
    - Format requests have a special background thread pool since that is highly latency sensitive (note that sync requests that are in front of the format request still have to complete first, but this means it wont compete with other background requests)
    - As a side note, it seems nice that the cli no longer needs tokio as a dependency, probably making the cli more portable and easier to build on more platforms?

- Drops tower-lsp, which is currently not very well maintained and has given us some issues in the past (which we have successfully worked around, but we knew eventually we'd have to switch away)

- Completely non-locking implementation. Background threads get a `DocumentSnapshot` (this could expand to include more things) containing an `Arc<TextDocument>` and `Arc<Settings>` that they are free to use without having to, say, lock a `Mutex`. If the main thread gets a Sync request/notif that needs to update a `TextDocument` or `Settings` while a background thread is working with it, the main thread will clone that `TextDocument` before modifying it. We should try extremely hard to avoid `Mutex`es in background threads going forward.

- More control over request/notification handling, including some nice ergonomics improvements where `Initialize` and `Initialized` are handled before you enter the actual main loop, since those are "one time" requests/notifs. We also have full control over shutdown/exit.

- A very solid code base to build from and reference going forward. The ruff code base itself seems inspired by rust-analyzer, so we are inheriting a combination of the two, and can reference both going forward as potential sources of inspiration. Both also use lsp-server, which is good news for us in terms of maintainability of that crate.

Some other improvements:

- Better testing! We now have a global test client that is used across unit tests. It is stored behind a `Mutex`, which forces unit tests that require the client to run sequentially, while still allowing all other tests to run in parallel without issue. You use `with_client(|client| { })` in a unit test to access it. You just have to be thoughtful to not do something in your unit test that puts the client in a state that affects other unit tests. If you need to destructively modify the client for a test, you create an integration test instead (see `tests/initialization.rs` which tests start up and shut down).
    - Additionally, because we only have 1 client per process and the client tests are forced to run sequentially, using `-- --nocapture` actually makes sense now even without forcing 1 thread!


LICENSE
```diff
@@ -19,3 +19,33 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
+The `server` crate is a fork of `ruff_server`:
+
+- URL: https://github.com/astral-sh/ruff/tree/main/crates/ruff_server
+- Commit: aa429b413f8a7de9eeee14f8e54fdcb3b199b7b7
+- License: MIT
+
+MIT License
+
+Copyright (c) 2022 Charles Marsh
+
```
DavisVaughan, line 34: I tried the idea of putting some kind of header info in each file but it was not maintainable with the amount of changes I was also making to those files (deleting some, merging others, completely rewriting some, etc)

crates/server/src/message.rs
```diff
@@ -0,0 +1,55 @@
+use anyhow::Context;
+use lsp_types::notification::Notification;
+use std::sync::OnceLock;
+
+use crate::server::ClientSender;
+
+static MESSENGER: OnceLock<ClientSender> = OnceLock::new();
```
DavisVaughan, line 7: In addition to `logging.rs`, this is the one other piece of global state we now have

This allows us to have free functions like `show_err_msg!()` that sends `ShowMessage` notifications to the client to pop up toast notifications for the user when something massive has gone wrong

crates/server/src/server/api/notifications/did_change.rs
```diff
@@ -0,0 +1,37 @@
...
+    ) -> Result<()> {
+        let key = session.key_from_url(uri);
+
+        session
+            .update_text_document(&key, content_changes, new_version)
+            .with_failure_code(ErrorCode::InternalError)?;
+
+        Ok(())
+    }
+}
```
DavisVaughan, line 37: This is what "implementing" a sync handler looks like. We implement a super trait of `NotificationHandler` that gets the notification type, and then a more specific `SyncNotificationHandler` (or `BackgroundNotificationHandler`) trait that actually handles the request/notif.

In this case, since this is a sync request we have full mutable access to the server `Session`, so we just forward the information on to `session.update_text_document()`

crates/server/src/server/api/requests/view_file.rs
```diff
@@ -0,0 +1,140 @@
...
+    type Params = ViewFileParams;
+    type Result = String;
+    const METHOD: &'static str = "air/viewFile";
+}
+
+impl super::RequestHandler for ViewFile {
+    type RequestType = ViewFile;
+}
+
+impl super::BackgroundDocumentRequestHandler for ViewFile {
```
DavisVaughan, line 43: This is what implementing a custom request looks like. We have to implement `lsp_types::request::Request` for our custom request and supply the params/result types, but it's pretty easy!

crates/server/src/server/client.rs
```diff
@@ -0,0 +1,169 @@
...
+            .into(),
+        )
+    }
+}
+
+impl<'s> Requester<'s> {
+    /// Sends a request of kind `R` to the client, with associated parameters.
+    /// The task provided by `response_handler` will be dispatched as soon as the response
+    /// comes back from the client.
+    pub(crate) fn request<R>(
```
DavisVaughan, line 98: Typically the server _receives_ requests/notifs from the client, but in a few places we will also _send_ requests/notifs to the client, possibly in the middle of some other request.

That is handled too! We have access to a `client.requester` that we can send along a request through. When the response comes back, a handler that we provided is immediately run. The response is matched up to our handler through a `RequestId` that is created during the request.

We use this to register dynamic capabilities, like `didChangeWatchedFiles` server support

crates/server/src/session.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Tracks what LSP features the client supports and doesn't support.
+    resolved_client_capabilities: ResolvedClientCapabilities,
+}
+
+/// An immutable snapshot of `Session` that references
+/// a specific document.
+pub(crate) struct DocumentSnapshot {
+    document_ref: index::DocumentQuery,
+    position_encoding: PositionEncoding,
+}
```
DavisVaughan, line 30: The `Session` is like our `LspState` and the `DocumentSnapshot` is like our `WorldState`

We can expand on these as needed going forward, but should be careful about what we put in `DocumentSnapshot`

crates/server/src/test/client.rs
```diff
@@ -0,0 +1,48 @@
...
+/// be able to use it, create an integration test instead, which runs in its own process.
+static TEST_CLIENT: LazyLock<Mutex<TestClient>> = LazyLock::new(|| Mutex::new(test_client()));
+
+pub(crate) fn with_client<F>(f: F)
+where
+    F: FnOnce(&mut server_test::TestClient),
+{
+    let mut client = TEST_CLIENT.lock().unwrap();
+    f(&mut client)
+}
```
DavisVaughan, line 30: Our global test client

DavisVaughan:
We'll probably revisit at some point in the future, but no need to keep this open

Issue #123, opened 2024-12-21
CRLF makes `--check` report a change while there is nothing to modify


etiennebacher:
I have two files:

1. `no_crlf.R`, created on Linux:
```r
if (TRUE) {
}
```
```
> file no_crlf.R
no_crlf.R: ASCII text
```
2. `with_crlf.R`, created on Windows and then opened on Linux:
```r
if (TRUE) {
}
```
```
> file with_crlf.R
with_crlf.R: ASCII text, with CRLF line terminators
```

Both files have identical contents and are already formatted, but (I guess that) the `CRLF` line terminators make `--check` report some changes to make:
```bash
> air format no_crlf.R && air format no_crlf.R --check
> air format with_crlf.R && air format with_crlf.R --check
Would reformat: <path>/with_crlf.R
```

---

Tested with `air 0.1.1`. 

Thanks again for this amazing tool

DavisVaughan:
Ah yes, that's a side effect of https://github.com/posit-dev/air/issues/90 and will get fixed when we fix that

Thanks for being an early beta tester! Nice to surface a few of these early bugs

Issue #90, opened 2024-12-10
Dont normalize line endings in the cli


DavisVaughan:
https://github.com/posit-dev/air/blob/68554e4b85528af9f3b33a15a654258750dbd159/crates/air/src/commands/format.rs#L188-L199

Will require `normalize_string()` in the formatter for multiline strings

https://github.com/rome/tools/pull/1672 is the old rome PR that allows this to work right. What i missed before is that on the way out of the formatter the printer normalizes all line endings one final time to whatever `LineEnding` is set as. So it is impossible to get mixed line endings on the way out.

PR #127, opened 2025-01-03
Don't normalize strings in the CLI

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/90
Closes https://github.com/posit-dev/air/issues/123

Follow up to https://github.com/posit-dev/air/pull/78

- The CLI now never normalizes line endings, allowing `--check` to work correctly, and allowing us to take advantage of an optimization where we detect that no changes occurred during the formatting
- The LSP continues to always normalizes line endings to Unix endings. Since _everything_ is Unix line endings there, we could add the optimization there too (and pre-emptively return `None` in the response to the format request if we detect no changes, rather than running an expensive diff algorithm)

As mentioned in https://github.com/posit-dev/air/issues/90, our parser and Biome's formatter are happy with alternative line endings when they appear in the _trivia_, but if non-unix line endings appear in a _token_ then the formatter panics. This happens in multiline strings, so we now normalize strings efficiently using a `Cow::Borrowed` when nothing changed.

The way this all flows through biome is (as implemented by https://github.com/rome/tools/pull/1672):
- We send a parse tree with CRLF line endings into the formatter
- The formatter normalizes all _trivia_ to Unix line endings
- We normalize all _tokens_ to Unix line endings (only multiline strings have this issue)
- At Print time, the formatter turns a Unix line ending into the user requested `LineEnding` 

The last step there is why the Printer doesn't allow any non-Unix line endings internally. It really just looks for `\n` at print time to decide when to apply `LineEnding`, so `\r\n` would make it behave incorrectly.

### DIFF
--- a/crates/air_r_formatter/src/string_literal.rs
+++ b/crates/air_r_formatter/src/string_literal.rs
@@ -6,7 +6,6 @@ use biome_formatter::trivia::format_replaced;
 use biome_formatter::Format;
 use biome_formatter::FormatResult;
 use std::borrow::Cow;
-use std::cell::Cell;
 
 use crate::context::RFormatContext;
 use crate::RFormatter;
@@ -36,7 +35,6 @@ impl<'token> FormatStringLiteralToken<'token> {
 
         let text = token.text_trimmed();
         let text = normalize_string(text);
-        let text = Cell::new(text);
 
         FormatNormalizedStringLiteralToken { token, text }
     }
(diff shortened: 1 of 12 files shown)


crates/air_r_formatter/src/string_literal.rs
```diff
@@ -0,0 +1,147 @@
...
+                self.text.take(),
+                self.token,
+                self.token.text_trimmed_range().start(),
+            ),
+        )
+        .fmt(f)
+    }
+}
+
+/// Normalize a string, returning a [`Cow::Borrowed`] if the input was already normalized
```
lionel-, line 74: Maybe mention why using the `line_ending` won't work here, because it mutates strings in place?

lionel- (approved):

crates/air_r_formatter/src/string_literal.rs
DavisVaughan: I also anticipate a small chance that we might do further "normalization" here, like using a consistent quote style or something similar

PR #130, opened 2025-01-07
Use a global test client for all unit tests

DavisVaughan:
A feature plucked from https://github.com/posit-dev/air/pull/126, and a step towards allowing dynamic setting of `logLevel` and `dependencyLogLevels` in the logging code, which are going to require some global state.

Here are notes from that other PR about this:

> We now have a global test client that is used across unit tests. It is stored behind a `Mutex`, which forces unit tests that require the client to run sequentially, while still allowing all other tests to run in parallel without issue. You use `with_client(|client| { })` in a unit test to access it. You just have to be thoughtful to not do something in your unit test that puts the client in a state that affects other unit tests. If you need to destructively modify the client for a test, you create an integration test instead (see `tests/initialization.rs` which tests start up and shut down). Additionally, because we only have 1 client per process and the client tests are forced to run sequentially, using `-- --nocapture` actually makes sense now even without forcing 1 thread!

The async code makes this quite tricky. `[#tokio::test]` creates 1 runtime _per test_, which completely screws you up if you have a global object that needs to be safely synchronized across tests, because you can't really synchronize across runtimes. The temporary solution used here is to use a global test runtime as well, and just call `block_on()` for each test. See https://github.com/tokio-rs/tokio/issues/2374 for more.

The better solution in the future will be to remove the async code in favor of lsp-server, which will simplify everything down to this: https://github.com/posit-dev/air/blob/feature/ruff/crates/server/src/test/client.rs

---

I've also added in the ability for our test client to send the `initialized()` notification. This will end up being required when we switch to lsp-server, as lsp-server waits for `initialized()` before letting us proceed on the server side.


crates/lsp/src/handlers.rs
```diff
@@ -46,10 +46,13 @@ pub(crate) async fn handle_initialized(
...
-        .register_capability(regs)
-        .instrument(span.exit())
-        .await?;
+    if !regs.is_empty() {
+        client
+            .register_capability(regs)
+            .instrument(span.exit())
+            .await?;
+    }
+
```
DavisVaughan, line 55: Now that we send `initialized()`, this `handle_initialized()` is fired and the server was sending this request to the client that the client test code was not expecting. Since the test client doesn't actually have the capability required here, I guarded it like this to avoid sending the request entirely.

(Note this was quite hard to debug, as it seems like the codec code can't figure out how to deserialize this message from the server at all)

crates/lsp/src/logging.rs
```diff
@@ -204,26 +204,18 @@ pub(crate) fn init_logging(
...
-    if is_test_client(client_info) {
-        // During parallel testing, `set_global_default()` gets called multiple times
-        // per process. That causes it to error, but we ignore this.
-        tracing::subscriber::set_global_default(subscriber).ok();
-    } else {
-        tracing::subscriber::set_global_default(subscriber)
-            .expect("Should be able to set the global subscriber exactly once.");
-    }
+    tracing::subscriber::set_global_default(subscriber)
+        .expect("Should be able to set the global subscriber exactly once.");
```
DavisVaughan, line 208: Yay

crates/lsp_test/src/lsp_client.rs
```diff
@@ -117,15 +122,18 @@ impl TestClient {
...
         // recognise the Shutdown request.
-        let req = Request::build("shutdown").id(self.id()).finish();
+        let req = Request::build("shutdown").id(id.clone()).finish();
 
         // Unwrap: For this test client it's fine to panic if we can't send
         self.tx.send(req).await.unwrap();
-        self.recv_response().await;
+
+        let response = self.recv_response().await;
+        assert_eq!(&id, response.id());
```
DavisVaughan, line 136: I switched `id()` to return `jsonrpc::Id` because I found it was quite useful to add in some extra asserts to ensure that a call to `recv_response()` actually returns the response you were hoping for

.github/workflows/test-linux.yml
```diff
@@ -54,5 +54,4 @@ jobs:
         env:
           AIR_LOG_LEVEL: trace
         # `--nocapture` to see our own `tracing` logs
-        # `--test-threads 1` to ensure `tracing` logs aren't interleaved
-        run: cargo test -- --nocapture --test-threads 1
+        run: cargo test -- --nocapture
```
lionel-, line 57: it's nice to make CI and local testing more consistent.

crates/lsp/src/handlers_format.rs
```diff
@@ -233,133 +233,132 @@ fn find_expression_lists(node: &RSyntaxNode, offset: TextSize, end: bool) -> Vec
...
+        })
     }
 
     // https://github.com/posit-dev/air/issues/61
-    #[tests_macros::lsp_test]
-    async fn test_format_minimal_diff() {
-        let mut client = init_test_client().await;
+    #[test]
+    fn test_format_minimal_diff() {
+        with_client(|client| async {
```
lionel-, line 260: A bit unfortunate to increase nesting but the explicitness is nice.

crates/lsp/src/test/client.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// If you need to modify a client/server pair in such a way that no other unit tests will
+/// be able to use it, create an integration test instead, which runs in its own process.
+///
+/// TODO: When we switch off async tower-lsp, make this a simpler `LazyLock` around a
+/// `std::sync::Mutex`, and then actually `lock()` in `with_client()`, only providing
+/// a `&mut TestClient` to the unit test.
+static TEST_CLIENT: OnceCell<Mutex<TestClient>> = OnceCell::const_new();
+
+/// `#[tokio::test]` creates a new runtime for each individual test. That doesn't work
```
lionel-, line 31: It will be nice to remove all these tokio complications.

lionel- (approved):
LG!

PR #136, opened 2025-01-08
Implement source of truth `settings` crate

### LINKED ISSUE #132: Can we have one "source of truth" for all air options?
Follow up to https://github.com/posit-dev/air/pull/122

Here are the current set of options:

- `IndentStyle`
- `IndentWidth`
- `LineEnding` (note, `Auto` option in TOML file)
- `LineLength`
- `MagicLineBreak` (note, boolean in TOML file)

Right now we are kind of in a weird place. We have our own shim types for the options that come from Biome because we want to add extra features or have different defaults. But if we change the defaults of these shim types in the `workspace` crate, that isn't reflected in our tests in `air_r_formatter`.

This makes me think we should explore having a single "source of truth" for each of these options.

I'm thinking that `air_r_formatter` should be the home for all of these options shims. Our `RFormatOptions` struct would use _our_ option types, and our implementation of biome's `FormatOptions` trait (like `indent_style() -> IndentStyle`) is where we'd convert to the Biome type for use within the guts of the formatter.

We'd go ahead and add serde capabilities here, for use in the workspace crate.

This would allow us to make `air_r_formatter` the source of truth for these options, and if we change the default there it propagates out everywhere (including tests).

The one exception would be `LineEnding`, since in the TOML file we support an additional `Auto` option that the formatter doesn't actually see (the formatter gets the resolved result of the auto option). We'd have 2 copies of `LineEnding`, one without `Auto` in `air_r_formatter`, and one with `Auto` in the workspace crate. This would be the way we'd handle any other options going forward that need a slightly different representation for use in the TOML file.

DavisVaughan:
Closes #132 

This ended up feeling _very_ good, I like this a lot


crates/air_r_formatter/src/context.rs
```diff
@@ -138,20 +138,20 @@ impl RFormatOptions {
...
-    fn line_width(&self) -> LineWidth {
-        self.line_width
+    fn line_width(&self) -> biome_formatter::LineWidth {
+        self.line_width.into()
     }
 
-    fn line_ending(&self) -> LineEnding {
-        self.line_ending
+    fn line_ending(&self) -> biome_formatter::LineEnding {
+        self.line_ending.into()
```
DavisVaughan, line 154: This is the only place we convert from our `settings::` types to `biome_formatter::` types. That's it! That's all the glue! Go traits!

crates/air_r_formatter/src/context.rs
```diff
@@ -8,16 +8,16 @@ use biome_formatter::BracketSpacing;
...
-use biome_formatter::IndentStyle;
-use biome_formatter::IndentWidth;
-use biome_formatter::LineEnding;
-use biome_formatter::LineWidth;
 use biome_formatter::TransformSourceMap;
+use settings::IndentStyle;
+use settings::IndentWidth;
+use settings::LineEnding;
+use settings::LineWidth;
+use settings::MagicLineBreak;
```
DavisVaughan, line 16: It's implicit here, but note that our `RFormatOptions` struct now uses our settings types

crates/air_r_formatter/tests/specs/r/binary_expression.R.snap
```diff
@@ -527,7 +527,7 @@ base_version <-
 
 -----
 Indent style: Tab
-Indent width: 2
+Indent width: 4
```
DavisVaughan, line 530: Ignore all the diffs here, nothing actually changes because the indent style was tab

In a follow up PR I'll move us over to 2 spaces as the default

crates/settings/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+license.workspace = true
+repository.workspace = true
+
+[features]
+biome = ["dep:biome_formatter"]
+serde = ["dep:serde"]
+
+[dependencies]
+biome_formatter = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
```
DavisVaughan, line 20: Note that serde support and biome conversion support are optional

Probably overkill but it feels quite "professional" to do it this way

crates/settings/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[dev-dependencies]
+anyhow = { workspace = true }
+biome_formatter = { workspace = true }
+insta = { workspace = true }
+serde = { workspace = true }
+# Self dependency to activate optional features for tests
+settings = { path = "./", features = ["biome", "serde"] }
```
DavisVaughan, line 28: This was a clever trick I copied from the jiff crate

crates/workspace/src/settings/line_ending.rs
DavisVaughan, line 1: Here is our 1 special setting that has extra options, Auto and Native, when specified in the TOML file. It gets "finalized" into a narrower `settings::LineEnding`

PR #139, opened 2025-01-09
Support dynamic updates of the log level

### LINKED ISSUE #134: Make `logLevel` and `dependencyLogLevels` dynamic
After https://github.com/posit-dev/air/pull/130, which would allow us to add global state in the logging system since we'd only have 1 test client per process. 

We will also need to "pull" configuration for `air.logLevel` and `air.dependencyLogLevels` when configuration changes, like we do for document config settings. This will also be the basis for how we pull any global or workspace level configuration in the future, with this being a global configuration setting.

DavisVaughan:
Closes #134 

https://github.com/user-attachments/assets/d1d284dd-8355-4dc0-a729-44951ee297f4

### DIFF
--- a/crates/lsp/src/logging.rs
+++ b/crates/lsp/src/logging.rs
@@ -15,9 +15,15 @@
 //!
 //! - The environment variable `AIR_LOG_LEVEL` is consulted.
 //!
-//! - The LSP `InitializeParams.initializationOptions.logLevel` option is consulted. This
-//!   can be set in VS Code / Positron using `air.logLevel`, or in Zed by supplying
-//!   `initialization_options`.
+//! - The configuration variable `air.logLevel` is consulted. This can be set in two ways:
+//!
+//!   - As an initialization option, i.e. `InitializeParams.initializationOptions.logLevel`,
+//!     which is passed in on startup. In VS Code / Positron our extension preemptively
+//!     looks for the global `air.logLevel` configuration option and passes it through.
+//!     In Zed you'd set this in `initialization_options`.
+//!
+//!   - As a dynamic global configuration option. We watch for `air.logLevel` configuration
+//!     changes and update the log level accordingly.
 //!
 //! - If neither are supplied, we fallback to `"info"`.
 //!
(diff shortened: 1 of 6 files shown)


crates/lsp/src/logging.rs
```diff
@@ -157,14 +157,67 @@ impl<'a> MakeWriter<'a> for LogWriterMaker {
...
+                self.log_level = log_level;
+                self.dependency_log_levels = dependency_log_levels;
+            }
+            Err(error) => {
+                // Log and return without updating our internal log level
+                tracing::error!("Failed to update log level: {error}");
+            }
+        }
+    }
+}
```
DavisVaughan, line 211: This is the core bit of it, pretty nice!

crates/lsp/src/main_loop.rs
```diff
@@ -165,6 +166,9 @@ pub(crate) struct LspState {
 
     /// List of client capabilities that we care about
     pub(crate) capabilities: ResolvedClientCapabilities,
+
+    /// State used to dynamically update the log level
+    pub(crate) log_state: Option<logging::LogState>,
```
DavisVaughan, line 171: I'm excited to move to lsp-server because I think the startup procedure there will result in us creating `GlobalState` / `LspState` _after_ we've received `InitializeParams`, meaning we will be able to get rid of `Option<>`s here and in `GlobalState`

crates/lsp/src/main_loop.rs
```diff
@@ -232,7 +237,7 @@ impl GlobalState {
         // tower-lsp backend and the Jupyter kernel.
         let (events_tx, events_rx) = tokio_unbounded_channel::<Event>();
 
-        let (log_state, log_tx) = LogState::new(client.clone());
+        let (log_thread_state, log_tx) = LogThreadState::new(client.clone());
```
DavisVaughan, line 240: Just a rename so we can use `LogState` for the bit that holds the reload handle

This thread state thing will also eventually go away in lsp-server world

lionel- (approved):

PR #146, opened 2025-01-12
Add Folding Range functionality to air

kv9898:
Currently I got an example folding range working, which is a good start for future work!

### DIFF
--- a/crates/lsp/src/folding_range.rs
+++ b/crates/lsp/src/folding_range.rs
@@ -1,0 +1,17 @@
+use tower_lsp::lsp_types::FoldingRange;
+use tower_lsp::lsp_types::FoldingRangeKind;
+
+use crate::documents::Document;
+
+pub fn folding_range(_document: &Document) -> anyhow::Result<Vec<FoldingRange>> {
+    // sample
+    let example_range = FoldingRange {
+        start_line: 0,
+        start_character: None,
+        end_line: 1,
+        end_character: None,
+        kind: Some(FoldingRangeKind::Region),
+        collapsed_text: None,
+    };
+    Ok(vec![example_range])
+}
--- a/crates/lsp/src/handlers.rs
+++ b/crates/lsp/src/handlers.rs
@@ -7,12 +7,16 @@
 
 use struct_field_names_as_array::FieldNamesAsArray;
 use tower_lsp::lsp_types;
+use tower_lsp::lsp_types::FoldingRange;
+use tower_lsp::lsp_types::FoldingRangeParams;
 use tower_lsp::Client;
 use tracing::Instrument;
 
 use crate::config::VscDiagnosticsConfig;
 use crate::config::VscDocumentConfig;
+use crate::folding_range::folding_range;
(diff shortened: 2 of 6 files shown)


kv9898:
I'm sorry that the current form is not based on biome-rowan, but I struggled with the fact that the rowan ast does not treat comments as equivalent elements in the tree (rather as children of the previous element?), and I didn't figure out how to extract the cursor location (line row and column), which are important for folding ranges. With tree-sitter, I can mimick a lot of existing AIR code and get it working quickly. Maybe we can talk about converting this tree-sitter-based protytype to biome_rowan later?

kv9898:
@lionel- @DavisVaughan Let me know your thoughts when you guys are less busy.

crates/lsp/src/folding_range.rs
```diff
@@ -0,0 +1,379 @@
...
+    region_marker: &mut Option<usize>,
+    cell_marker: &mut Option<usize>,
+) {
+    let node = cursor.node();
+    let _field_name = match cursor.field_name() {
+        Some(name) => format!("{name}: "),
+        None => String::new(),
+    };
+
+    let start = node.start_position();
```
kv9898, line 62: I couldn't find an equivalent of this using biome_rowan

crates/lsp/src/folding_range.rs
```diff
@@ -0,0 +1,379 @@
...
+            let folding_range = bracket_range(
+                start.row,
+                start.column,
+                end.row,
+                end.column - 1,
+                count_leading_whitespaces(document, end.row),
+            );
+            folding_ranges.push(folding_range);
+        }
+        "comment" => {
```
kv9898, line 81: biome_rowan treats comments as children of the next command. This is super annoying. Things are much neater with treesitter

crates/lsp/src/folding_range.rs
```diff
@@ -0,0 +1,379 @@
...
+        Some(name) => format!("{name}: "),
+        None => String::new(),
+    };
+
+    let start = node.start_position();
+    let end = node.end_position();
+    let node_type = node.kind();
+
+    match node_type {
+        "parameters" | "arguments" | "braced_expression" => {
```
kv9898, line 67: Let me know if I miss some of other node types with brackets that I should take care of

kv9898:
@lionel- any updates?

DavisVaughan:
We are focusing a ton of time just trying to get the initial version of the formatter out the door, we will come back and review this in detail after that, and we really appreciate your patience, sorry it is taking so long, it's just a very bit context switch for us at the moment!

kv9898:
@DavisVaughan @lionel- seems like air is now shipped in Positron. Is it now time to come back to this PR?

DavisVaughan:
> this PR was originally for ark anyway and I was told to move it to here

Yea we apologize for that. We now think we can promptly get your PR reviewed and merged if it is on the Ark side, since it is "just one more LSP feature for Ark", rather than "the first non-formatting LSP feature for Air" which has much higher overhead for us to consider.

Issue #140, opened 2025-01-09
Add a tracing subscriber in the CLI


DavisVaughan:
Probably a simple one that emits to stderr, but we need to be able to see those logs!

PR #155, opened 2025-01-14
Add global `--log-level` and `--no-color` CLI options and logging support

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/140

The cli help file:

Screenshot 2025-01-14 at 3 34 21 PM

Running air format on R source results in a few failures that we see now

Screenshot 2025-01-14 at 3 33 39 PM

### DIFF
--- a/crates/air/src/args.rs
+++ b/crates/air/src/args.rs
@@ -2,6 +2,8 @@ use clap::Parser;
 use clap::Subcommand;
 use std::path::PathBuf;
 
+use crate::logging;
+
 #[derive(Parser)]
 #[command(
     author,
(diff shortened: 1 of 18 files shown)


crates/air/src/args.rs
```diff
@@ -39,3 +44,18 @@ pub(crate) struct FormatCommand {
...
+pub(crate) struct GlobalOptions {
+    /// The log level. One of: `error`, `warn`, `info`, `debug`, or `trace`. Defaults
+    /// to `warn`.
+    #[arg(long, global = true)]
+    pub(crate) log_level: Option<logging::LogLevel>,
+
+    /// Disable colored output. To turn colored output off, either set this option or set
+    /// the environment variable `NO_COLOR` to any non-zero value.
+    #[arg(long, global = true)]
+    pub(crate) no_color: bool,
```
lionel-, line 60: I'd maybe move this to an advanced section in the future. Since we don't have many options it'd be weird to do this now.

lionel- (approved):

PR #158, opened 2025-01-14
Document current stance on `air.toml`s

DavisVaughan:
In response to #148


lionel- (approved):
I added a commit to mention setting synchronisation with the IDE. Can you review it please?

Issue #56, opened 2024-11-25
Blockers for alpha release


DavisVaughan:
- [x] Decide on defaults for formatter
- [x] Add some basic documentation
- [x] Add the TOML configuration parser (it will be nice to have hierarchical configuration option setting eventually) https://github.com/posit-dev/air/issues/83 (#122)
- [x] Develop the CLI enough to be able to format with a few configurable options (#122)
- [x] Binaries for air using ruff approach https://github.com/posit-dev/air/pull/67
- [ ] OpenVSX extension (https://github.com/posit-dev/air/pull/166)
- [ ] VS Code Marketplace extension (https://github.com/posit-dev/air/pull/166)
- [x] README with installation instructions 
    - [x] Instructions for Positron / VS Code
    - [x] Instructions for RStudio

ntluong95:
Great tools! I have tried to install the CLI in my computer and it worked

PR #166, opened 2025-01-15
Add support for locating a bundled version of `air`

DavisVaughan:
- [x] Merge after #160 since it will have some merge conflicts
- [x] Switch to `"bundled"` by default
- [x] Figure out prettier trailing commas

Closes https://github.com/posit-dev/air/issues/56

This PR goes with https://github.com/posit-dev/air/pull/164. Together it gives us the ability to release the Air extension to OpenVSX and the Code Marketplace _with_ the air binary bundled in. It works for macOS (x86 and arm) and Windows (x86) right now.

A manual workflow run against this PR resulted in https://github.com/posit-dev/air/actions/runs/12796811126 with the artifacts you see at the bottom there. The Windows one worked out of the box on my Windows machine without having air installed! The Mac one works for me locally, I set `air.executableLocation: "bundled"` to easily test it.

---

I'm very open to ideas to improve on this one. No strong feelings on the naming of the new `air.executableLocation` setting, but I do think we should probably default to `"environment"` so by default we look in the `PATH` and fallback to the bundled version if we can't find anything there. That should provide a more consistent experience when going from the command line to your IDE.

It's quite easy to bundle a binary alongside your VS Code extension. You just place the binary in the extension folder and it gets slurped up. In the GitHub Action merged in https://github.com/posit-dev/air/pull/164 I placed it in `bundled/bin/<air/air.exe>`. This PR adds support for finding that bundled binary.

I know we discussed not bundling for the first release but it didn't seem too hard and feels like it would just be a much nicer experience overall.

### DIFF
--- a/editors/code/src/constants.ts
+++ b/editors/code/src/constants.ts
@@ -4,6 +4,7 @@ const folderName = path.basename(__dirname);
 
 /**
  * Path to the root directory of this extension.
+ * https://github.com/microsoft/vscode-python-tools-extension-template/blob/main/src/common/constants.ts
  */
 export const EXTENSION_ROOT_DIR =
 	folderName === "common"
--- a/editors/code/src/constants.ts
+++ b/editors/code/src/constants.ts
@@ -24,5 +24,5 @@ export const BUNDLED_AIR_EXECUTABLE = path.join(
 	EXTENSION_ROOT_DIR,
 	"bundled",
 	"bin",
-	AIR_BINARY_NAME
+	AIR_BINARY_NAME,
 );
--- a/editors/code/src/constants.ts
+++ b/editors/code/src/constants.ts
@@ -1,0 +1,27 @@
+import * as path from "path";
+
+const folderName = path.basename(__dirname);
+
+/**
+ * Path to the root directory of this extension.
+ */
+export const EXTENSION_ROOT_DIR =
+	folderName === "common"
+		? path.dirname(path.dirname(__dirname))
+		: path.dirname(__dirname);
+
+/**
+ * Name of the `air` binary based on the current platform.
+ */
+export const AIR_BINARY_NAME = process.platform === "win32" ? "air.exe" : "air";
+
+/**
+ * Path to the `air` executable that is bundled with the extension.
+ * The GitHub Action is in charge of placing the executable here.
+ */
+export const BUNDLED_AIR_EXECUTABLE = path.join(
+	EXTENSION_ROOT_DIR,
+	"bundled",
+	"bin",
+	AIR_BINARY_NAME
+);
--- a/editors/code/src/logging.ts
+++ b/editors/code/src/logging.ts
@@ -1,0 +1,28 @@
+import * as util from "util";
+import { Disposable, OutputChannel } from "vscode";
+
+type Arguments = unknown[];
+class OutputChannelLogger {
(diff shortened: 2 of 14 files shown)


editors/code/src/constants.ts
```diff
@@ -0,0 +1,27 @@
+import * as path from "path";
+
+const folderName = path.basename(__dirname);
+
+/**
+ * Path to the root directory of this extension.
+ */
+export const EXTENSION_ROOT_DIR =
```
DavisVaughan, line 8: Comes from https://github.com/microsoft/vscode-python-tools-extension-template/blob/main/src/common/constants.ts

editors/code/src/logging.ts
DavisVaughan, line 1: This is a global logger giving us access to an `outputLog()` free function we can call from anywhere

You set it with `registerLogger()` and set it to the output channel that the server also ends up using.

editors/code/src/workspace.ts
```diff
@@ -0,0 +1,51 @@
+import path from "path";
+import * as vscode from "vscode";
+import * as fs from "fs-extra";
+
+export async function getRootWorkspaceFolder(): Promise<vscode.WorkspaceFolder> {
```
DavisVaughan, line 5: This also comes from https://github.com/microsoft/vscode-python-tools-extension-template/blob/8f474ec4ac4e7205ffed9f7f243473bb00bf29c0/src/common/utilities.ts#L38

It seems reasonable?

editors/code/src/logging.ts
```diff
@@ -0,0 +1,28 @@
...
+	channel = new OutputChannelLogger(logChannel);
+	return {
+		dispose: () => {
+			channel = undefined;
+		},
+	};
+}
+
+export function outputLog(...args: Arguments): void {
+	if (process.env.CI === "true") {
```
lionel-, line 24: Nice

editors/code/src/logging.ts
```diff
@@ -0,0 +1,28 @@
...
+export function registerLogger(logChannel: OutputChannel): Disposable {
+	channel = new OutputChannelLogger(logChannel);
+	return {
+		dispose: () => {
+			channel = undefined;
+		},
+	};
+}
+
+export function outputLog(...args: Arguments): void {
```
lionel-, line 23: Can the file be called `output` and this function `log()` so that we'd do `output.log()`?

editors/code/src/lsp.ts
```diff
@@ -89,7 +99,7 @@ export class Lsp {
 
 						const config = vscode.workspace.getConfiguration(
 							undefined,
-							{ uri, languageId },
+							{ uri, languageId }
```
lionel-, line 102: oops, commas are disappearing :P

editors/code/src/workspace.ts
lionel-: For some definition of reasonable ;)
I mean I would expect all this complexity to be hidden behind a VS Code API.

lionel- (approved):
Looks good. My main comment is that I think we should favour the bundled executable.

- The formatter will eventually become very stable, so the chance of inconsistencies between the CLI and the LSP will be slim.

- The bulk of Air features in the future will be in the LSP. And there will likely be several integrations that require the extension to match the executable. It would be very easy for a user to unknowingly get stuck with a CLI executable and not know they need to manually update it regularly.

PR #174, opened 2025-01-16
Polish `package.json` a bit

DavisVaughan:
Some of these end up showing up on the installed extension's landing page

See also https://code.visualstudio.com/api/references/extension-manifest

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -1,16 +1,28 @@
 {
 	"name": "air",
 	"displayName": "Air",
-	"description": "Air Language Server",
+	"description": "R formatter and language server for Visual Studio Code",
 	"version": "0.0.1",
 	"publisher": "Posit",
 	"license": "MIT",
-	"repository": "https://github.com/posit-dev/air",
+	"homepage": "https://github.com/posit-dev/air",
+	"repository": {
+		"type": "git",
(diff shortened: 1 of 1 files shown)


editors/code/package.json
```diff
@@ -1,16 +1,28 @@
 {
 	"name": "air",
 	"displayName": "Air",
-	"description": "Air Language Server",
+	"description": "R formatter and language server for Visual Studio Code",
 	"version": "0.0.1",
 	"publisher": "Posit",
 	"license": "MIT",
-	"repository": "https://github.com/posit-dev/air",
+	"homepage": "https://github.com/posit-dev/air",
```
juliasilge, line 8: What would you think about using the docs site URL here instead?

DavisVaughan: Good idea!

editors/code/package.json
```diff
@@ -1,16 +1,28 @@
 {
 	"name": "air",
 	"displayName": "Air",
-	"description": "Air Language Server",
+	"description": "R formatter and language server for Visual Studio Code",
```
lionel-, line 4: Since that will show up in Positron too

```suggestion
	"description": "R formatter and language server",
```

lionel- (approved):

PR #175, opened 2025-01-16
Activate extension immediately

DavisVaughan:
Rather than waiting for an R file to open, @lionel- what do you think?


lionel- (approved):
Sounds good!

PR #192, opened 2025-01-21
docs: Repeatable install commands

schloerke:
* Use `(<code>)` which calls a subshell to run the code, allowing `cd` to not change the callers working directory.
* Use `|| true` as the first time ran, there are no files to remove and `zsh` throws an error. (I saw SO answers saying `-f` should resolve the error I'm seeing, but it did not work as is. I'm happy to adjust this if there is a better code option.)
* Use `npx @vscode/vsce` to auto install `vsce`. When calling `npx vsce` directly, a warning is displayed asking to upgrade to `@vscode/vsce`.

### DIFF
--- a/.github/workflows/release-vscode.yml
+++ b/.github/workflows/release-vscode.yml
@@ -68,7 +68,7 @@ jobs:
 
       # Build the extension.
       - name: Package Extension (release)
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Thanks!

PR #198, opened 2025-01-22
Add intro to formatter section

lionel-:


docs/formatter.qmd
```diff
@@ -6,4 +6,180 @@ editor:
...
+
+1 + 2 + 3
+```
+
+The goal of this feature is to strike a balance between being opionated and being flexible.
+
+# When does a formatter run?
+
+The most convenient way to run a formatter is when you save a file.
+This is not typically enabled by default so you will need to enable it manually (see the [Editors](editors.qmd) section).
```
DavisVaughan, line 176:
```suggestion
Most editors don't enable this by default, so check out the [Editors](editors.qmd) section to learn how to configure your editor to format on save.
```

docs/formatter.qmd
```diff
@@ -6,4 +6,180 @@ editor:
...
+Once enabled, your code will be reformatted every time you save!
+This prevents your code from ever diverging from the standards set by Air.
+Another way of running the formatter in an IDE is when you paste code (this is enabled by the setting "Format on paste").
```
DavisVaughan, line 179: I think I would leave this sentence out for now, since I feel like the main point of this section is:
- Learn how to format on save
- Learn that the CLI exists to format a whole project at once

docs/formatter.qmd
```diff
@@ -6,4 +6,180 @@ editor:
...
+
+Air also provides a [CLI tool](https://posit-dev.github.io/air/cli.html).
+It makes it easy to reformat whole directories such as an R project or package.
```
DavisVaughan, line 182: > If you're incorporating Air into an existing project, you may want to use the CLI tool to run `air format` once on the entire project, commit those results, and then rely on per-file format on save from there on out.

Maybe another sentence like this?

docs/formatter.qmd
```diff
@@ -6,4 +6,180 @@ editor:
...
+
+Note that in all of these cases, the formatter will only run if your code is grammatically correct.
```
DavisVaughan, line 184:
```suggestion
Note that in all of these cases, the formatter will only run if your code is syntactically correct.
```

? 

In my head I tend to tie "grammatically" to "human language" so this confused me a little, but I don't mind too much if you want to keep it

docs/formatter.qmd
```diff
@@ -6,4 +6,180 @@ editor:
...
+If the formatter is seemingly no longer working, check your code for any syntax error.
```
DavisVaughan, line 185:
```suggestion
If the formatter is seemingly no longer working, check your code for syntax errors.
```

DavisVaughan (approved):
I think this looks great! It'll be very nice to have this on the website, and then it should be easy to transition to the Discussion post with minimal effort

docs/formatter.qmd
lionel-: hmm... that makes sense for the website. For the disucssions page, it's probably still be useful in order to collect feedback on that feature.

Issue #194, opened 2025-01-21
"bundled" version isn't compatible with Intel Macs (arm64 executable)


DivadNojnarg:
I installed air on 2 different machines, a Mac M2 (a week ago) and an imac intel (today). For the latter, I had to do extra steps to make it start, described below.

The error message I had

```
air Language Server client: couldn't create connection to server.

Using bundled executable as requested by `air.executableLocation`: /Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/bundled/bin/air
[Error - 21:43:41] Air Language Server client: couldn't create connection to server.
Error: spawn Unknown system error -86
	at ChildProcess.spawn (node:internal/child_process:421:11)
	at Object.spawn (node:child_process:777:9)
	at /Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:350928
	at async S.createConnection (/Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:183353)
	at async S.start (/Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:173807)
	at async t.Lsp.startImpl (/Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:93568)
	at async /Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:92347
	at async /Users/davidgranjon/.positron/extensions/posit.air-vscode-0.2.0-darwin-x64/dist/extension.js:1:441412
```

---- 

For the Intel Mac this is what I have:

```shell
R version 4.2.2 Patched (2022-10-31 r83219)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Ventura 13.7.2
```

with Positron:

```shell
Positron Version: 2025.01.0 (Universal) build 152
Code - OSS Version: 1.95.0
Commit: 66aa3fb7f98eb8d19155cb7220856154f6ede8b3
Date: 2025-01-06T02:47:02.050Z
Electron: 32.2.1
Chromium: 128.0.6613.186
Node.js: 20.18.0
V8: 12.8.374.38-electron.0
OS: Darwin x64 22.6.0
```

I installed the cli 0.2.0 and downloaded the Air - R Language Support extension from the market place.

[...]

PR #201, opened 2025-01-23
Switch to manually downloading the binary

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/194

In theory we could use `arch -x86_64 curl` on Mac and keep using cargo-dist, but due to https://github.com/axodotdev/cargo-dist/issues/1720 this does not work. I'm not really sure https://github.com/axodotdev/cargo-dist/issues/1720 will get solved, as when I think about this more we are kind of abusing cargo-dist here. It's purpose is really to get you set up to _run_ the tool (so bypassing Rosetta does seem to make sense on their end), all we were using it for is a way to magically download the right one for this platform, but that isn't too hard to set up ourself.

This dry run shows this seems to be working right
https://github.com/posit-dev/air/actions/runs/12935503617

I downloaded the x86 mac vsix artifact it created and `file` shows that the air binary bundled in there is indeed the right architecture now. The output from the run on Windows shows that the `air.exe` still makes it into the `bundled/bin` directory correctly.

### DIFF
--- a/.github/workflows/release-vscode.yml
+++ b/.github/workflows/release-vscode.yml
@@ -45,9 +45,9 @@ jobs:
         with:
           fetch-depth: ${{ env.FETCH_DEPTH }}
 
-      # macOs or Linux binary
-      # Download, unpack, move into place, and clean up
-      - if: ${{ startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu') }}
+      - name: Download air binary (macOS or Linux)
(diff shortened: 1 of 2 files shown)


.github/workflows/release-vscode.yml
```diff
@@ -72,7 +82,7 @@ jobs:
 
       # Upload the extension.
       - name: Upload artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
DavisVaughan, line 85: I was forced to bump this by github actions

lionel- (approved):

PR #204, opened 2025-01-25
Switch to nextest for testing

### LINKED ISSUE #154: Seriously look into nextest
https://nexte.st/

I did a POC locally and was very impressed

We have had a long history of having to deal with tests that rely on global state. Since `cargo test` runs them in parallel within the same process, we often need some painful locking mechanisms which are annoying for us to manage.

The big selling point of nextest is that it is fast while giving _each test its own process_
https://nexte.st/docs/design/why-process-per-test/

That seems like it would work perfectly for us, so we should seriously look into this, and if it works well consider backporting to ark, as it might help with stability issues there

DavisVaughan:
Closes #154 

Switch to https://nexte.st/ for testing instead of `cargo test`

This allows us to drastically simplify our testing infrastructure, as test mode now more closely mimics "real life" which is guaranteed to only have 1 instance of Air per process. Big benefits of this include:

- No need to manage global state, so the LSP global client and tokio runtime were removed
- No need to clean up between tests (think, objects defined in R, on the ark side, or opened files in the air lsp)
- No need for integration tests. Everything already runs in its own process, which was typically the main benefit of an integration test. It also means we get access to non-public API for these tests again (because they are standard unit tests), and we should be able to use `cfg(test)` as needed in our code (but hopefully we won't need to anymore!).

I added a new developer doc section on how to get set up with nextest. Should just be installing the cli tool and then using `just test` or `just test-insta`.

I think we should consider sponsoring https://github.com/sunshowers


.config/nextest.toml
DavisVaughan, line 1: Values in this file are CI recommendations as mentioned in https://nexte.st/docs/configuration/

.github/workflows/test-linux.yml
```diff
@@ -50,8 +55,8 @@ jobs:
...
         run: cargo build
 
-      - name: Run unit tests
+      - name: Run tests
         env:
+          NEXTEST_PROFILE: "ci"
           AIR_LOG_LEVEL: trace
-        # `--nocapture` to see our own `tracing` logs
-        run: cargo test -- --nocapture
+        run: cargo nextest run
```
DavisVaughan, line 62: nextest will show all stdout and stderr for any failing tests, so I don't think we really need `-- --nocapture` anymore.

The nextest equivalent is `--no-capture`, but that also (smartly) forces sequential tests to prevent live output from interleaving between tests and I'd rather not sacrifice parallelism just to see output on passing tests.

crates/air/src/commands/format.rs
```diff
@@ -255,3 +255,50 @@ impl Display for FormatCommandError {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
```
DavisVaughan, line 260: These moved from integration tests back to being unit tests. And they can also now live in the same file (previously the global logger was preventing that)

crates/lsp/src/handlers_format.rs
```diff
@@ -244,132 +244,125 @@ fn find_expression_lists(node: &RSyntaxNode, offset: TextSize, end: bool) -> Vec
...
-    use crate::{documents::Document, test::with_client, test::TestClientExt};
+    use crate::{documents::Document, test::new_test_client, test::TestClientExt};
 
-    #[test]
-    fn test_format() {
-        with_client(|client| async {
-            let mut client = client.lock().await;
+    #[tokio::test]
+    async fn test_format() {
+        let mut client = new_test_client().await;
```
DavisVaughan, line 252: The diff here is painful, but the simplification is quite nice:
- Can now create a `new_test_client()` in every test
- Test itself gets marked as a tokio test (until we can remove tokio)
- `with_client()` is removed, flattening a level of nesting

This should transfer nicely to lsp-server as well, we just remove the awaits, async, etc.

crates/lsp/src/test/client.rs
DavisVaughan, line 1: SOOOOO much nicer

Even with lsp-server we were going to need a global `TEST_CLIENT` to prevent logging from being set up more than once per test process. This is so much better.

crates/lsp/src/test/client.rs
```diff
@@ -70,3 +15,55 @@ async fn new_test_client() -> TestClient {
 
     client
 }
+
+#[cfg(test)]
```
DavisVaughan, line 19: Again, can move this out of integration tests into unit tests

PR #205, opened 2025-01-25
Build for ARM Windows

### LINKED ISSUE #170: Add support for `"aarch64-pc-windows-msvc"`
After #166 

Hopefully is fairly easy. Requires updating the github action for `release-vscode.yml` and the cargo dist toml file. Hopefully everything else just works. If not, we can leave it for later.

DavisVaughan:
Closes #170 

Successful dry-run built an ARM Windows binary here https://github.com/posit-dev/air/actions/runs/12969634669/job/36173931978, in theory it works? 😆 

The way this works is through cross-compilation. GitHub currently only provides x86_64 Windows runners on the free tier, so instead according to https://github.com/axodotdev/cargo-dist/pull/1635 cargo-dist will detect that you are requesting ARM Windows and will automatically:
- Use x86_64-linux-gnu Rust on an ubuntu-20.04 runner
- Install the `messense/cargo-xwin` image which brings in https://github.com/rust-cross/cargo-xwin for cross compilation to windows from linux (here is the docker image, its by the owner of cargo-xwin https://hub.docker.com/r/messense/cargo-xwin)

I see two reasons this could change on the cargo-dist side going forward:
- https://github.com/axodotdev/cargo-dist/issues/1644 could get implemented, where they use a simpler approach of using GitHub's x86_64 windows machine and cross compile to `aarch64-pc-windows-msvc`, which supposedly just works with `cargo build --target aarch64-pc-windows-msvc` (Biome actually does this)
- GitHub could release their ARM Windows runners under the free tier, like they recently did with ARM Linux runners. At which point cross compilation would go away entirely

Regardless, both possibilities would be cargo-dist details and would be nicely abstracted away from us, so all we'd have to do is update cargo-dist to get the changes. And in the meantime as long as cargo-xwin works I'm not mad about using it.


DavisVaughan:
This was mildly complicated by this issue https://github.com/axodotdev/cargo-dist/issues/1633

The summary of the problem was:
- `xtask/codegen` relies on ureq, which brings in rusttls, which brings in `ring`
- `ring` is hard to cross compile on ARM Windows using cargo-xwin, as noted in this issue https://github.com/rust-cross/cargo-xwin/issues/76, because of this line which forces usage of the `clang` compiler over `clang-cl` https://github.com/briansmith/ring/blob/a9b88826e7826a6f9422680fb6240f75511cf2c7/build.rs#L580, and that caused our build to fail

The easy solution was to just tell dist to only worry about managing `crates/air` by updating the workspace members in the dist toml, and by setting `precise-builds = true`. We should have been setting this in the first place, and that also shaves a whole minute off our build steps, as now its no longer trying to compiler heavier internal crates like `lsp-test` and `xtask/codegen`.

If we do end up having to bring rusttls (and therefore ring) back into the core stack before cargo-dist gets off xwin using one of the two means outlined above, then we will have to either set the new `XWIN_CROSS_COMPILER` environment variable, or ask cargo-dist to provide a way for us to set `--cross-compiler` (https://github.com/rust-cross/cargo-xwin?tab=readme-ov-file#customization). But I'm hoping that cargo-dist instead just moves off xwin in favor of just cross compiling straight from x86_64 windows runners, which seems to work fine.

lionel- (approved):

PR #209, opened 2025-01-31
Add tests for VS Code extension

### LINKED ISSUE #207: Create extension-level unit and integration tests
We currently test air from Rust with both unit tests and protocol-level tests with a dummy client. However we should also have integration tests in the VS Code extension:

- Some features are provided by the extension and should be tested on their own, such as settings synchronisation.

- We should also add some integration tests for major features. These will also be run by the Positron test suite to make sure everything is in working order. This will be important to ensure quality as Air increasingly becomes part of the R development experience in Positron.

lionel-:
Closes #207.

Adds VS Code extension tests for:

- Formatting with forward propagation of settings (from `.vscode/settings.json`).
- Formatting with backward propagation of settings (from `air.toml`)

These tests are run in a new CI job that is triggered when extension sources change.

We now also gracefully handle the case where the "use bundled binary" setting is set to true (the default) but there is no bundled binary. In that case we fall back to finding a binary on the `PATH`. This should be helpful for people running the extension on platforms for which we are not able to bundle a binary and this makes it easy to set up CI by installing Air on the path.


editors/code/src/binary.ts
```diff
@@ -1,5 +1,7 @@
 import * as vscode from "vscode";
```
lionel-, line 1: We now check for a binary on the path if there is none bundled.

editors/code/src/extension.ts
```diff
@@ -11,6 +11,13 @@ export function activate(context: vscode.ExtensionContext) {
 
 	ctx = new Ctx(context, lsp);
 	registerCommands(ctx);
+
+	return {
```
lionel-, line 15: This is a way to allow other extensions to call an API provided by Air. In this case we are exporting a private API for our own unit tests so they can access state in the Air instance.

editors/code/src/lsp.ts
```diff
@@ -44,6 +51,15 @@ export class Lsp {
 		return this.client;
 	}
 
+	public waitForSettingsNotification(): Promise<void> {
```
lionel-, line 57: This provides a way to `await` the event.

editors/code/src/test/editor-utils.ts
```diff
@@ -0,0 +1,100 @@
+// From testUtils.ts in the typescript-language-feature extension
```
lionel-, line 1: Mostly from positron-r.

editors/code/src/test/extension.ts
```diff
@@ -0,0 +1,65 @@
...
+	// started up.
+	await internalApi().ctx.lsp.start();
+}
+
+export async function withToml(
+	content: string,
+	run: () => Promise<void>,
+): Promise<void> {
+	const createdToml: Promise<void> =
+		internalApi().ctx.lsp.waitForSettingsNotification();
```
lionel-, line 54: This is the trick to ensure the test is synchronised with the LSP.

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,51 @@
+name: "Test / Code extension"
+
+on:
+  push:
+    branches:
+      - main
+    paths: ["editors/code/**"]
+  pull_request:
+    paths: ["editors/code/**"]
+  workflow_dispatch:
```
DavisVaughan, line 10: Nice yea I was going to request workflow-dispatch in case we change the server and want to run it manually

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Report rust toolchain
+        run: rustup show
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Setup SSH access
+        uses: mxschmitt/action-tmate@v3
+        if: ${{ inputs.ssh }}
+        timeout-minutes: 30
```
DavisVaughan, line 34: Never used, remove?

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Install X Virtual Framebuffer
+        run: |
+          sudo apt-get install xvfb
+
+      - name: Run tests
+        env:
+          AIR_LOG_LEVEL: trace
+        run: |
+          export PATH="~/.cargo/bin/:$PATH"
```
DavisVaughan, line 48: I'd be surprised if we actually needed this since `/.cargo/bin` will already be on the path

.github/workflows/test-code.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: ${{ inputs.ssh }}
+        timeout-minutes: 30
+
+      - name: Install Air
+        run: |
+          cargo install --path "crates/air"
+
+      - name: Install X Virtual Framebuffer
+        run: |
+          sudo apt-get install xvfb
```
DavisVaughan, line 42: What's this?

editors/code/.vscode/settings.json
```diff
@@ -9,5 +9,6 @@
 		"dist": true // set this to false to include "dist" folder in search results
 	},
 	// Turn off tsc task auto detection since we have the necessary tasks as npm scripts
-	"typescript.tsc.autoDetect": "off"
+	"typescript.tsc.autoDetect": "off",
+	"tabWidth": 4
```
DavisVaughan, line 13: That's going to apply to all languages, like, even R files probably when you open a new one? Can we scope this to a language or something? I'm also not sure why we need this, the prettier config file should be working instead. You should not need the Zed config either unless something else is wrong.

editors/code/src/binary.ts
```diff
@@ -15,22 +17,38 @@ export async function resolveAirBinaryPath(
...
-			output.log(`Using bundled executable: ${BUNDLED_AIR_EXECUTABLE}`);
+		if (fs.existsSync(BUNDLED_AIR_EXECUTABLE)) {
+			output.log(
+				`Using bundled executable as requested by \`air.executableLocation\`: ${BUNDLED_AIR_EXECUTABLE}`,
+			);
 			return BUNDLED_AIR_EXECUTABLE;
 		}
+
+		// Fallthrough
+		output.log(`Bundled executable not found: ${BUNDLED_AIR_EXECUTABLE}`);
```
DavisVaughan, line 30:
```suggestion
		output.log(`Bundled executable not found, falling back to environment: ${BUNDLED_AIR_EXECUTABLE}`);
```

Makes the flow a little more self explanatory I think

editors/code/src/lsp.ts
```diff
@@ -30,11 +34,17 @@ export class Lsp {
 
 	private fileSettings: FileSettingsState;
 
+	public onSettingsNotificationEmitter: vscode.EventEmitter<void>;
```
DavisVaughan, line 37: Place the `Emitter` near the event itself?

editors/code/src/lsp.ts
```diff
@@ -30,11 +34,17 @@ export class Lsp {
...
+
 	constructor(context: vscode.ExtensionContext) {
 		this.channel = vscode.window.createOutputChannel("Air Language Server");
 		context.subscriptions.push(this.channel, registerLogger(this.channel));
+
 		this.stateQueue = new PQueue({ concurrency: 1 });
 		this.fileSettings = new FileSettingsState(context);
+
+		this.onSettingsNotificationEmitter = new vscode.EventEmitter<void>();
```
DavisVaughan, line 46: It looks like `vscode.EventEmitter` has a `dispose()` method. Should it be in `context.subscriptions.push()`?

editors/code/src/lsp.ts
```diff
@@ -136,9 +155,11 @@ export class Lsp {
...
 			clientOptions,
 		);
-		client.onNotification(SYNC_FILE_SETTINGS, (settings) =>
-			this.fileSettings.handleSettingsNotification(settings),
-		);
+
+		client.onNotification(SYNC_FILE_SETTINGS, (settings) => {
+			this.fileSettings.handleSettingsNotification(settings);
+			this.onSettingsNotificationEmitter.fire();
+		});
```
DavisVaughan, line 162:
```suggestion
		client.onNotification(SYNC_FILE_SETTINGS, (settings) => {
			this.onSettingsNotificationEmitter.fire(settings);
		});
```

What if we went a step further and made this the official way of notifying when `Lsp` receives a settings notification? We could make it so we call `this.fileSettings.handleSettingsNotification()` when `onSettingsNotification` fires. i.e. remove this line and change our constructor to

```ts
	constructor(context: vscode.ExtensionContext) {
		this.channel = vscode.window.createOutputChannel("Air Language Server");
		context.subscriptions.push(this.channel, registerLogger(this.channel));

		this.onSettingsNotificationEmitter =
			new vscode.EventEmitter<SyncFileSettingsParams>();
		this.onSettingsNotification = this.onSettingsNotificationEmitter.event;

		this.stateQueue = new PQueue({ concurrency: 1 });
		this.fileSettings = new FileSettingsState(context);

		context.subscriptions.push(
			this.onSettingsNotification((settings) => {
				this.fileSettings.handleSettingsNotification(settings);
			}),
		);
	}
```

And these would get typed

```ts
public onSettingsNotification: vscode.Event<SyncFileSettingsParams>;
public onSettingsNotificationEmitter: vscode.EventEmitter<SyncFileSettingsParams>;
```

editors/code/src/test/format.test.ts
```diff
@@ -0,0 +1,67 @@
...
+			assert.strictEqual(editor.options.insertSpaces, true);
+			assert.strictEqual(editor.options.indentSize, 8);
+			assert.strictEqual(editor.options.tabSize, 8);
+
+			await vscode.commands.executeCommand(
+				"editor.action.formatDocument",
+			);
+			await doc.save();
+
+			assert.strictEqual(doc.getText(), old);
```
DavisVaughan, line 37: I'm not sure I feel like this test is strong enough. It doesn't test that anything actually happened. i.e. 

```
await vscode.commands.executeCommand(
  "editor.action.formatDocument",
);
```

this could be a no-op and the test would pass, which makes me uncomfortable. I feel like we need two sets of files?
- One folder of fixtures, i.e. `fixtures/format.R`, and we format these and then read their contents
- One folder of snapshots, i.e. `snapshots/format.R` and we compare the formatted contents above to these contents

I don't think we need to auto generate the snapshot files like testthat does, we can manually make the before/after versions of `format.R` ourselves, but I do feel it is important to test that we are actually getting invoked, forcing _something_ to happen.

DavisVaughan (approved):
One main comment about `format.test.ts` but otherwise looks good!

.github/workflows/test-code.yml
lionel-: AFAIU this simulates a windowing environment. Needed to instantiate a VS Code app outside of an X server (the environment that supports GUIs on Linux).

editors/code/.vscode/settings.json
lionel-: I think it was needed in json files or something like that. Using the same tab width across all languages in Air seems like a feature to me?

editors/code/src/lsp.ts
lionel-: hmm this should actually be `private` and I think we keep private fields at the bottom?

editors/code/src/test/format.test.ts
lionel-: Good point. I'm now stripping whitespace from the input files. If the formatter is a no-op and didn't repair the missing whitespace the tests will now fail.

.github/workflows/test-code.yml
DavisVaughan: cargo itself

PR #223, opened 2025-02-13
Add support for `# fmt: skip` directives

### LINKED ISSUE #52: Add suppression comment support

DavisVaughan:
Closes #52 

Motivating examples, until we add "official" support for these with `fmt: tabular` or `fmt: align-right` or something:

```r
# fmt: skip
list(
  this      = 1,
  that      = 2,
  thisthing = 3,
  thatthing = 4,
  andthis   = 5
)

# fmt: skip
tribble(
  ~a, ~b,
   1,  2
)
```

Note that format directives are applied at the _node_ level, meaning they are applied to whatever node the comment gets attached to (as determined by `place_comment()` in the formatter). If a node is skipped, all of its children are also skipped.

@lionel- I'm definitely open to more tests cases if you can think of any

@lionel- I went ahead and added a changelog bullet and some docs on `# fmt: skip file`, it felt weird to talk about just `# fmt skip` knowing we are about to add this too. I've also gone ahead and added `FormatDirective::SkipFile`. In the end it felt like a separate directive than `FormatDirective::Skip` rather than an argument to that one. I'd also be open to `# fmt: skip-file` if you like that better.

### DIFF
--- a/crates/comments/src/directive.rs
+++ b/crates/comments/src/directive.rs
@@ -1,0 +1,96 @@
+#[derive(Debug, PartialEq)]
+pub enum Directive {
+    Format(FormatDirective),
+}
+
+#[derive(Debug, PartialEq)]
+pub enum FormatDirective {
+    Skip,
+    SkipFile,
+}
+
+/// Parse a comment directive
+///
+/// These take the form:
+///
+/// ```text
+/// # <category>: <command> <optional-argument>
+/// ```
+///
+/// Such as:
+///
+/// ```text
+/// # fmt: skip
+/// # fmt: skip file
+/// # fmt: tabular
+/// # fmt: align-right
+/// # lint: skip
+/// # lint: skip rule
+/// ```
+///
+/// Note that directives are applied to the node they are attached to.
+pub fn parse_comment_directive(text: &str) -> Option<Directive> {
+    let text = text.strip_prefix('#')?;
+    let text = text.trim_start();
+    let (category, text) = text.split_once(':')?;
+    let text = text.trim();
+
+    match category {
+        "fmt" => parse_format_directive(text),
+        _ => None,
+    }
+}
+
+#[inline]
+fn parse_format_directive(text: &str) -> Option<Directive> {
+    match text {
+        "skip" => Some(Directive::Format(FormatDirective::Skip)),
+        "skip file" => Some(Directive::Format(FormatDirective::SkipFile)),
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+mod test {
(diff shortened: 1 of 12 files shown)


docs/formatter.qmd
```diff
@@ -173,6 +174,48 @@ list(foo, bar)
...
+# fmt: skip
+tribble(
+  ~x, ~y,
+   1,  2,
+)
+```
+
+`# fmt: skip file` skips the entire file.
+This must be placed at the top of the file before the first non-comment R code, otherwise the comment is ignored.
+Use this for generated files, or as an alternative to `exclude` for individual files that you know you'd like to exclude from formatting.
```
etiennebacher, line 202: Very happy to see this land in `air`! I was just taking a look at how this is implemented and noticed something unclear to me in the docs.

Is the sentence about generated files actually useful for end users? It seems to me that people who develop the framework should include this in the generated files, but end users shouldn't since it will be overwritten next time the file is generated. The example following this paragraph is also a bit confusing in this regard.

crates/comments/src/directive.rs
```diff
@@ -0,0 +1,96 @@
...
+/// # fmt: skip file
+/// # fmt: tabular
+/// # fmt: align-right
+/// # lint: skip
+/// # lint: skip rule
+/// ```
+///
+/// Note that directives are applied to the node they are attached to.
+pub fn parse_comment_directive(text: &str) -> Option<Directive> {
+    let text = text.strip_prefix('#')?;
```
lionel-, line 33: This needs to be `trim_start_matches('#')` for e.g. `## fmt: skip file`.

I've done this in the downstream PR.

docs/formatter.qmd
lionel-, line 202: To address Étienne's suggestion:

```suggestion
This is useful for generated files, or as an alternative to `exclude` for individual files that you know you'd like to exclude from formatting.
```

docs/formatter.qmd
```diff
@@ -173,6 +174,48 @@ list(foo, bar)
...
+
+``` r
+# Generated file - Don't modify by hand!
```
lionel-, line 205: Maybe we should mention the Generated by: convention?

lionel- (approved):

crates/comments/src/directive.rs
DavisVaughan: I actually purposefully did not do that to ensure that something that comes through here must start with `#`. i.e. `trim_start_matches('#')` would be fine with `fmt: skip file` without any `#`

PR #228, opened 2025-02-18
Fix handling of hugging calls

### LINKED ISSUE #21: Adding hugging in `fn1(fn2())` style calls
```r
abort(glue::glue(
  "Length implied by `dim`, {n_elements}, must match the length of `x`, {n_x}."
))
```

Formatted as this right now

```r
abort(
  glue::glue(
    "Length implied by `dim`, {n_elements}, must match the length of `x`, {n_x}."
  )
)
```

But its quite nice with the hug

(Something about enabling this when the only unnamed argument to a function call is another function call, but should respect that leading line break if user wants to expand!)

(quite common across dplyr and elmer)

lionel-:
Branched from #227 (for access to `Either` type).
Closes #21.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
   ))
   ```
 
-  are no longer expanded (#21).
+  are no longer fully expanded (#21).
 
 - Assigned pipelines no longer double-indent (#220).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@
 
 - Assigned pipelines no longer double-indent when a persistent line break is used (#220).
 
+- Hugging calls like:
+
+  ```r
+  list(c(
+    1,
+    2
+  ))
+  ```
+
+  are no longer expanded (#21).
+
+- Assigned pipelines no longer double-indent (#220).
+
 
 # 0.3.0
 
--- a/crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
@@ -76,6 +76,326 @@ impl RCallLikeArguments {
             Self::Subset2(node) => node.syntax(),
         }
     }
+
+    /// Writes the function arguments
+    ///
+    /// The "grouped" argument is either the first or last argument depending on the
+    /// `group_layout`, but currently it is always the last one.
+    ///
+    /// - If any arguments that aren't the grouped argument *force* a break, then we
+    ///   print in fully expanded mode.
+    ///
+    /// - If the grouped argument is an inline function with `parameters` that would
+    ///   *force* a break, then we print in fully expanded mode. We only want to
+    ///   allow forced breaks in a braced expression body.
+    ///
+    /// If neither of those trigger fully expanded mode, we best-fit between three
+    /// possible forms:
+    ///
+    /// ## Most expanded
+    ///
+    /// The `(`, `)`, and all arguments are within a single `group()`, and that
+    /// group is marked with `should_expand(true)`. The arguments are wrapped in
+    /// `soft_block_indent()`, and each argument is separated by a
+    /// `soft_line_break_or_space()`. Due to the forced expansion, these all
+    /// become hard indents / line breaks, i.e. the "most expanded" form.
+    ///
+    /// Example:
+    ///
+    /// ```r
+    /// map(
+    ///   xs,
+    ///   function(x) {
+    ///     x + 1
+    ///   }
+    /// )
+    /// ```
+    ///
+    /// ## Most flat
+    ///
+    /// Arguments are not grouped, each argument is separated by a
+    /// `soft_line_break_or_space()`, no forced expansion is done.
+    ///
+    /// Special formatting is done for a grouped argument that is an inline
+    /// function. We remove any soft line breaks in the `parameters`, which
+    /// practically means the only place it is allowed to break is in the function
+    /// body (but the break is not forced).
+    ///
+    /// Example:
+    ///
+    /// ```r
+    /// # NOTE: Not currently possible, as the `{}` always force a break right now,
+    /// # but this would be an example if `{}` didn't force a break.
+    /// map(xs, function(x) {})
+    /// ```
+    ///
+    /// This variant is removed from the set if we detect that the grouped argument
+    /// contains a forced break in the body (if a forced break is found in the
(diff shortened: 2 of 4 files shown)


crates/air_r_formatter/tests/specs/r/call.R
```diff
@@ -641,3 +641,67 @@ fn(
...
+c(list(foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz))
+
+# Call: Hugging case
+c(list(foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz()))
+
+# Another hugging case
+c(list(foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz(
+    1,
+    2
+)))
```
DavisVaughan, line 666:
```suggestion
# Call: Recursive hugging case, no breaks
c(list(foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz()))

# Call: Recursive hugging case, inner arguments break
c(list(foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz(1, 2)))

# Call: Recursive hugging case, persistent newlines
c(list(foobar(
  1,
  2
)))
```

For the 2nd example here, I think it is useful to start with `(1, 2)` flat so you can check how it expands it

And then I think we should add the 3rd case above to check that those persistent newlines kick in in the recursive case as they should

crates/air_r_formatter/tests/specs/r/call.R
```diff
@@ -641,3 +641,67 @@ fn(
...
+
```
DavisVaughan, line 667: Here's one more

```r
# Call: Recursive hugging case, persistent line breaks stop hugging
c(list(
  foobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbafoobarbazzzzzzzzzfoobarbaz(1, 2)))
```

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -944,20 +945,53 @@ impl Format<RFormatContext> for FormatAllArgsBrokenOut<'_> {
...
+    // We only consider calls with one argument
+    let Some(item) = items.first() else {
+        return Ok(false);
+    };
+    if items.len() != 1 {
+        return Ok(false);
+    }
+
+    // Unwrap the value to get the `AnyRExpression`
+    let Some(arg) = item.element().node.clone()?.value() else {
```
DavisVaughan, line 978:
```suggestion
    let Some(arg) = item.element().node()?.value() else {
```

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+- Hugging calls like:
+
+  ```r
+  list(c(
+    1,
+    2
+  ))
+  ```
+
+  are no longer expanded (#21).
```
DavisVaughan, line 14:
```suggestion
  are no longer fully expanded (#21).
```

DavisVaughan (approved):

lionel-:
oh I just noticed that these test cases are actually different, and the second one currently fails!

```r
c(list(
    1
) #foo
)

c(list(
  1
)
#foo
)
```

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
lionel-: The clone is necessary

lionel-: I can make it work with:

```rust
    let Some(arg) = item.element().node.as_ref().ok() else {
        return Ok(false);
    };

    let Some(arg) = arg.value() else {
        return Ok(false);
    };
```

but I'm not sure that's a win

lionel-:
I had to refactor things a bit to give the formatting method access to the syntax tree so I could check for comments a bit more flexibly. See last commit, should hopefully be reasonable.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -76,6 +76,326 @@ impl RCallLikeArguments {
             Self::Subset2(node) => node.syntax(),
         }
     }
+
```
lionel-, line 79: This is now a method on `RCallLikeArguments`, hence the big diff as the whole thing moved. It feels right and this gives access to `self` and the syntax elements.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -76,6 +76,326 @@ impl RCallLikeArguments {
...
+        // First write the most expanded variant because it needs `arguments`.
+        let most_expanded = {
+            let mut buffer = VecBuffer::new(f.state_mut());
+            buffer.write_element(FormatElement::Tag(Tag::StartEntry))?;
+
+            write!(
+                buffer,
+                [FormatAllArgsBrokenOut {
+                    call: self,
```
lionel-, line 253: We can then pass `self` here.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -919,6 +932,7 @@ impl Format<RFormatContext> for FormatGroupedArgument {
 }
 
 struct FormatAllArgsBrokenOut<'a> {
+    call: &'a RCallLikeArguments,
```
lionel-, line 935: The whole call is now part of this struct. This way we can peek at anything we want if needed.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -945,7 +959,7 @@ impl Format<RFormatContext> for FormatAllArgsBrokenOut<'_> {
             Ok(())
         });
 
-        let args = if !self.expand && is_hugging_call(self.args)? {
+        let args = if !self.expand && is_hugging_call(self.args, self.call.r_token())? {
```
lionel-, line 962: Taking a peek at the right delimiter so we can check for comments.

PR #236, opened 2025-02-19
Add a space in complex anonymous functions

### LINKED ISSUE #235: Single sided `~` formulas with non-identifier RHS should add a space
I think we mentioned just keeping things simple here and never adding the space (instead encouraging anonymous functions), but it adds enough noise in dplyr diffs that I feel like we should probably just do it. I don't think it will be that hard. It's also in the style guide

Image

Image

DavisVaughan:
Closes #235

### DIFF
--- a/crates/air_r_formatter/src/r/auxiliary/unary_expression.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/unary_expression.rs
@@ -20,10 +20,11 @@ impl FormatNodeRule<RUnaryExpression> for FormatRUnaryExpression {
             if f.comments().has_comments(argument.syntax())
                 && !f.comments().is_suppressed(argument.syntax())
             {
-                // We never line break between the operator and its argument except if an
-                // existing comment separates them
+                // If an existing comment separates the operator and its argument, we are
+                // forced to line break
                 write!(f, [soft_block_indent(&argument.format())])
             } else {
+                // Otherwise we never line break
                 write!(f, [argument.format()])
             }
         });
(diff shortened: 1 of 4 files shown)


crates/air_r_formatter/src/r/auxiliary/unary_expression.rs
```diff
@@ -11,17 +14,37 @@ impl FormatNodeRule<RUnaryExpression> for FormatRUnaryExpression {
...
-        }
+        let argument = format_with(|f| {
+            if f.comments().has_comments(argument.syntax())
+                && !f.comments().is_suppressed(argument.syntax())
+            {
+                // We never line break between the operator and its argument except if an
+                // existing comment separates them
+                write!(f, [soft_block_indent(&argument.format())])
+            } else {
+                write!(f, [argument.format()])
```
lionel-, line 27: This comment confused me at first because it mentions the behaviour in the other branch

```suggestion
                // If an existing comment separates the operator and its
                // argument we are forced to line break
                write!(f, [soft_block_indent(&argument.format())])
            } else {
                // Otherwise we never line break
                write!(f, [argument.format()])
```

crates/air_r_formatter/src/r/auxiliary/unary_expression.rs
```diff
@@ -11,17 +14,37 @@ impl FormatNodeRule<RUnaryExpression> for FormatRUnaryExpression {
...
+/// We define "complex" as anything except a simple identifier, so even `~ "foo"` results
+/// in a space.
+fn is_complex_anonymous_function(operator: &RSyntaxToken, argument: &AnyRExpression) -> bool {
+    // Must be a `~`
+    if !matches!(operator.kind(), RSyntaxKind::TILDE) {
+        return false;
+    }
+
+    // "Complex" is anything except identifiers
+    !matches!(argument, AnyRExpression::RIdentifier(_))
```
lionel-, line 49: I would expect `~NA` and `~NULL` to be treated like `~foo`?
And probably other literals like `~10` too?

lionel- (approved):
Do we have tests for chained unary formulas?

It should look like:

```r
~~foo
~~ 1 + 2
```

PR #240, opened 2025-02-19
Skip semicolons as whitespace when not between two tokens

### LINKED ISSUE #238: Trailing `;` at the end of the file causes a parser panic
Minimal reprex

```r
1;
```

This is the cause of the styler panic

This is related to us not _actually_ handling semicolons and just having them magically accidentally fall out of the parse tree

DavisVaughan:
Closes #238 

And adds a whole bunch of extra tests, we didn't have any before


lionel- (approved):
Tricky!

PR #244, opened 2025-02-20
Add support for skipping a file

### LINKED ISSUE #219: Ignore auto generated files and an option to include auto generated files
> I think there is a general consensus to ignore files with keywords like `@generated`.
> 
> So that the Cargo.lock files also contain it.
> 
> https://github.com/posit-dev/air/blob/16d3b53a7258106ab28a40d2e47573913d445dad/Cargo.lock#L1 

 _Originally posted by @eitsupi in [#128](https://github.com/posit-dev/air/issues/128#issuecomment-2610076285)_

For example
- rustfmt: <https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#format_generated_files>
- Go: <https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source>

Any file that contains a `@generated` comment marker within the first 5 lines of the file is considered auto-generated and excluded, but can optionally include these auto-generated files.

In the absence of such a thing, we would need to maintain a list like the one included in #213, which I believe is not sustainable.

lionel-:
Branched from #223.

Closes #219.


crates/air_r_formatter/src/lib.rs
```diff
@@ -209,6 +209,10 @@ where
         f.context().comments().is_suppressed(node.syntax())
     }
 
+    fn fmt_suppressed(&self, node: &N, f: &mut RFormatter) -> FormatResult<()> {
```
DavisVaughan, line 212:
```suggestion
    /// Formats a suppressed node
    ///
    /// You may want to override this method if you need to manually handle
    /// formatting of a suppressed node. This should be extremely rare.
    fn fmt_suppressed(&self, node: &N, f: &mut RFormatter) -> FormatResult<()> {
```

To match the other `fmt_*()` methods with defaults in the trait here

crates/air_r_formatter/src/r/auxiliary/root.rs
```diff
@@ -23,4 +40,58 @@ impl FormatNodeRule<RRoot> for FormatRRoot {
...
+        let Some(child) = node.expressions().first() else {
+            return false;
+        };
+
+        let comments = f.context().comments();
+
+        // Only consider very first comment in the file
+        let Some(comment) = comments.leading_comments(child.syntax()).first() else {
+            return false;
+        };
```
DavisVaughan, line 54: Hmm one of my examples was actually

```
# Generated file - Don't modify by hand!
# fmt: skip file
```

while that may end up working for other reasons, the point was that `# fmt: skip file` could be anywhere at the top as long as it is before the first actual R node (so, there could be other leading comments before it)

I'm not sure what you think about that

crates/air_r_formatter/src/r/auxiliary/root.rs
```diff
@@ -1,10 +1,27 @@
...
 pub(crate) struct FormatRRoot;
+
+impl FormatRRoot {
+    fn is_suppression_comment(text: &str) -> bool {
+        matches!(
+            parse_comment_directive(text),
+            Some(Directive::Format(FormatDirective::SkipFile))
+        )
+    }
+}
```
DavisVaughan, line 23: I think I would just inline this at this point?

crates/air_r_formatter/src/r/auxiliary/root.rs
```diff
@@ -23,4 +40,58 @@ impl FormatNodeRule<RRoot> for FormatRRoot {
...
+
+        Self::is_suppression_comment(comment.piece().text())
+    }
+
+    // The default handling of suppressed nodes in `FormatNodeRule::fmt()`
+    // does not work out well for `RRoot` because it doesn't format
+    // trivia verbatim. So we use a custom implementation instead.
+    fn fmt_suppressed(&self, node: &RRoot, f: &mut RFormatter) -> FormatResult<()> {
+        let RRootFields { bom_token, .. } = node.as_fields();
+        write!(f, [bom_token.format()])?;
```
DavisVaughan, line 64: IIUC you should not need this? You recursively format `node` and all its descendants, and since `node` is RRoot and the bom token is a descendant of that, it should get formatted too i think?

crates/air_r_formatter/src/r/auxiliary/root.rs
```diff
@@ -23,4 +40,58 @@ impl FormatNodeRule<RRoot> for FormatRRoot {
             ]
         )
     }
+
+    fn is_suppressed(&self, node: &RRoot, f: &RFormatter) -> bool {
+        let Some(child) = node.expressions().first() else {
+            return false;
+        };
+
+        let comments = f.context().comments();
```
DavisVaughan, line 49:
```suggestion
        let comments = f.context().comments();

        comments.mark_suppression_checked(node.syntax());

        let Some(child) = node.expressions().first() else {
            return false;
        };
```

Even though the immediate call to `fmt_suppressed()` will recursively mark everything as suppressed, the correct thing to do here in the overrride of `is_suppressed()` is to also call `mark_suppression_checked()` on this node, since you are declaring that you are checking its suppression manually.

For an example, see TsUnionType in Biome

```rs
    fn is_suppressed(&self, node: &TsUnionType, f: &JsFormatter) -> bool {
        f.comments().mark_suppression_checked(node.syntax());

        if node.types().is_empty() {
            f.comments().is_suppressed(node.syntax())
        } else {
            // Suppression applies to first variant
            false
        }
    }
```

crates/air_r_formatter/src/r/auxiliary/root.rs
```diff
@@ -23,4 +40,58 @@ impl FormatNodeRule<RRoot> for FormatRRoot {
...
+        write!(f, [bom_token.format()])?;
+
+        // Mark everything as visited so the formatter doesn't panic
+        // thinking we've missed elements. We're going to print each element
+        // manually with `dynamic_text()`.
+        mark_visited(node.syntax(), f);
+
+        // `node.text()` trims trivia so we get the raw text on the syntax
+        // node instead
+        let text: String = node.syntax().text().into();
```
DavisVaughan, line 73: I wonder how expensive this folding of individual text pieces ends up being. Like, for very large generated files it does kind of stink that we have to waste time and actually do work on that file to "format" it, vs just recognizing the file as generated and not touching it whatsoever

crates/comments/src/directive.rs
```diff
@@ -30,15 +30,21 @@ pub enum FormatDirective {
 ///
 /// Note that directives are applied to the node they are attached to.
 pub fn parse_comment_directive(text: &str) -> Option<Directive> {
-    let text = text.strip_prefix('#')?;
+    let text = text.trim_start_matches('#');
```
DavisVaughan, line 33:
```suggestion
    let text = text.strip_prefix('#')?;
    let text = text.trim_start_matches('#');
```

I propose this to enforce that at least 1 `#` must be present in `text`

crates/comments/src/directive.rs
```diff
@@ -60,9 +89,6 @@ mod test {
         let format_skip = Some(Directive::Format(crate::FormatDirective::Skip));
         let format_skip_file = Some(Directive::Format(crate::FormatDirective::SkipFile));
 
-        // Must have leading `#`
-        assert!(parse_comment_directive("fmt: skip").is_none());
```
DavisVaughan, line 64: I think this is important

crates/comments/src/directive.rs
```diff
@@ -30,15 +30,21 @@ pub enum FormatDirective {
...
+
+        #[allow(clippy::single_match)]
+        match category {
+            "fmt" => return parse_format_directive(text),
+            _ => {} // fallthrough
+        }
     }
+
+    let text = text.trim();
+    parse_special_directive(text)
```
DavisVaughan, line 47: This feels a bit too awkward, I'm not really sure I understand the purpose of the fallthrough

I think this allows weirdness like `# weird: Generated by`

I feel like we can probably simplify this somehow?

Maybe it's

```
let text = text.strip_prefix('#')?;
let text = text.trim_start_matches('#');
let text = text.trim_start();

if let Some(directive) = parse_special_directive(text) {
  return Some(directive);
}

let (category, text) = text.split_once(':')?;
let text = text.trim();

match category {
        "fmt" => parse_format_directive(text),
        _ => None,
}
```

crates/comments/src/directive.rs
```diff
@@ -50,6 +56,29 @@ fn parse_format_directive(text: &str) -> Option<Directive> {
...
+fn parse_special_skip_file(text: &str) -> Option<Directive> {
+    // Convention used in the R community (Roxygen, Rcpp, etc)
+    if text.starts_with("Generated by") {
+        return Some(Directive::Format(FormatDirective::SkipFile));
+    }
+
+    // Convention used in some programming language communities
+    if text.contains("@generated") {
+        return Some(Directive::Format(FormatDirective::SkipFile));
+    }
```
DavisVaughan, line 77: I would vote against doing this unless the community actively requested this from us, I would like to keep this feature as tightly scoped as possible

crates/comments/src/directive.rs
```diff
@@ -92,5 +121,11 @@ mod test {
             parse_comment_directive("#  fmt:  skip file"),
             format_skip_file
         );
+
+        assert_eq!(parse_comment_directive("# Generated by:"), format_skip_file);
+        assert_eq!(
+            parse_comment_directive("# foo\n# @generated\n # bar"),
```
DavisVaughan, line 127: This example doesn't make sense to me because we would never get more than 1 comment as a slice of `text` coming through here

DavisVaughan (approved):

crates/air_r_formatter/src/r/auxiliary/root.rs
lionel-: Sounds good to me

crates/air_r_formatter/src/r/auxiliary/root.rs
lionel-: It did use to be more complex

crates/air_r_formatter/src/r/auxiliary/root.rs
lionel-: We can add an optimisation to the CLI by handling skip-file comments even before parsing happens. We've decided to do that only if it proves necessary though.

crates/comments/src/directive.rs
lionel-: We've cleared this up one-on-one

crates/comments/src/directive.rs
lionel-: I've also added a comment with the example usage you shared with me yesterday.

crates/air_r_formatter/src/r/auxiliary/root.rs
lionel-: Removed it then had to add it back

Issue #243, opened 2025-02-20
Developer tool - `air.executableLocation` option of `"path"`, and new `air.executablePath`


DavisVaughan:
`air.executablePath` is an open text string, you insert the path to an air binary. This path is only considered when `air.executableLocation` is set to `"path"`

The point is so that we as developers can set `air.executablePath` to point at our debug binary location, and then toggle it on and off easily using `air.executableLocation`.

Currently with `"environment"` we have to do PATH management to avoid conflicts between release Air and debug Air, and that is quite annoying

DavisVaughan:
Plan:

- Rename `air.executableLocation` to `air.executableStrategy` and update documentation
- Add `"path"` option to `air.executableStrategy`
- Add `air.executablePath`

PR #245, opened 2025-02-20
Rename `air.executableLocation` and add `air.executablePath` support

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/243

- Renamed `air.executableLocation` to `air.executableStrategy` and updated documentation
- Added `"path"` option to `air.executableStrategy`
- Added `air.executablePath`

Reworked `resolveAirBinaryPath()` to be a series of if/else statements so it is _extremely clear_ what all of the fallback behavior combination is. I think this dramatically improves both readability and the logging. I used Rust style `Option`s by adding helpers like `airBinaryFromBundled()` that return `string | undefined`, which I think made the main function easier to read

I tested out each of these behaviors locally and am happy with the logging and errors

### DIFF
--- a/editors/code/CHANGELOG.md
+++ b/editors/code/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## Development version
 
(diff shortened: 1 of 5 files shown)


lionel- (approved):
Thanks!

PR #270, opened 2025-02-27
docs(neovim): add configuration guide for Neovim as LSP client

PMassicotte:
Introduce a new documentation file for configuring Neovim as a Language Server Protocol (LSP) client using nvim-lspconfig and conform.nvim. This guide provides step-by-step instructions for setting up Air as an LSP client and enabling automatic formatting on save. Update the main editors documentation to include a link to the new Neovim guide, expanding the list of supported editors.


m-muecke:
> Did that rapidly, please let me know if I missed something. This configuration has worked fine for me for the past few days.

Nice! The config for using nvim-lspconfig is missing and a guide for disabling the formatting compatibility of r-language-server when both are in use.

PMassicotte:
Let me try just calling setup.

PMassicotte:
You

> Nice! Since air is already merged in nvim-lspconfig one doesn't have to create a custom setup and the following is sufficient:
> 
> ```lua
> require("lspconfig").air.setup({})
> ```

You are absolutely right. I will modify accordingly.

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+
+First, install the Air client by following the instructions on the [Air CLI documentation](https://posit-dev.github.io/air/cli.html).
+
+# Configuration
+
+Air can be configured as a Language Server Protocol (LSP) client via [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) or as a formatter plugin via [conform.nvim](https://github.com/stevearc/conform.nvim).
+
+## LSP Configuration with nvim-lspconfig
+
+To configure Air as an LSP client, add the following to your `nvim/lua/plugins/lspconfig.lua`:
```
TymekDev, line 19: Personally, I would leave out the paths.
```suggestion
To configure Air's language server, add the following to your config:
```
It doesn't have to be that specific file—anywhere works,

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+
+```lua
+require("lspconfig").air.setup({})
+```
+
+To enable automatic formatting on save, add the following to your `nvim/after/plugin/lsp.lua` (or other appropriate location):
```
TymekDev, line 25:
```suggestion
To format files with Air on save, add the following to your configuration:
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+
+```lua
+vim.api.nvim_create_autocmd("BufWritePre", {
```
TymekDev, line 28: Do you think it makes sense to provide this autocmd? I'd expect everyone to already have something like this in their config.

However, for the completeness (e.g. for Neovim newcomers) it might make sense to keep it.

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+    pattern = { "*.R", "*.r" },
+    callback = function()
+        vim.lsp.buf.format({
+            async = true,
```
TymekDev, line 32:
```suggestion
```

I'd argue to leave that out given this note:

> Editing the buffer while formatting asynchronous can lead to unexpected changes.

I don't think it's necessary and—from what I tested—air is really quick at formatting.

docs/editor-neovim.qmd
TymekDev: Worth to consider: using `ftplugin`.

Putting the below command in either `ftplugin/r.lua` or `after/ftplugin/r.lua` would have a roughly the same effect. I supposed it's a style preference of using `pattern` vs. filetypes.
```lua
vim.api.nvim_create_autocmd("BufWritePre", {
  buffer = 0,
  callback = function()
    vim.lsp.buf.format()
  end,
})
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+            notify_on_error = true,
+            format_on_save = {
+                lsp_fallback = true,
+            },
+            formatters_by_ft = {
+                r = { "air" },
+            },
+        })
+    end,
+}
```
TymekDev, line 62: That's a `lazy.nvim`-specific setup. I'd suggest using a generic setup (example below) and point to conform's README for a way to install the plugin.

```lua
require("conform").setup({
  formatters_by_ft = {
    r = { "air" },
  },
})
```

What about filetypes for R markdown and Quatro?

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,63 @@
...
+                return client.name == "air"
+            end,
+        })
+    end,
+})
+```
+
+## conform.nvim configuration
+
+Air can be configured as a formatter plugin via [conform.nvim](https://github.com/stevearc/conform.nvim). This allows you to format R code with Air inside code chunks in Quarto and RMarkdown documents. It can be configured by adding the following to your `nvim/lua/plugins/conform.lua`:
```
TymekDev, line 43:
```suggestion
Air can be used as a formatter via [conform.nvim](https://github.com/stevearc/conform.nvim). This allows using Air to format R code inside code chunks in Quarto and RMarkdown documents. It can be configured by adding the following to your configuration:
```

> This allows you to format R code with Air inside code chunks in Quarto and RMarkdown documents.

Is this an out-of-the-box functionality or does one need to configure tree-sitter injections? 👀

TymekDev (changes requested):
Thanks for putting this together! I have added some comments with my feedback.

docs/editor-neovim.qmd
PMassicotte: Currently not supported by air.

docs/editor-neovim.qmd
PMassicotte: I think it is out of the box, my conform config is pretty simple.

docs/editor-neovim.qmd
PMassicotte: > Worth to consider: using `ftplugin`.
> 
> Putting the below command in either `ftplugin/r.lua` or `after/ftplugin/r.lua` would have a roughly the same effect. I supposed it's a style preference of using `pattern` vs. filetypes.
> 
> ```lua
> vim.api.nvim_create_autocmd("BufWritePre", {
>   buffer = 0,
>   callback = function()
>     vim.lsp.buf.format()
>   end,
> })
> ```

I think you are right, either are fine. Maybe we can propose both solution/approach.

docs/editor-neovim.qmd
DavisVaughan: @TymekDev and @PMassicotte can you please double check that Air works in various Quarto and RMarkdown documents? I have my doubts because I went looking in conform and could not find any specific Quarto support for LSP/formatting passthrough.

If it does seem to be working, I'd love to know what I missed so I can study how it works on the conform side

m-muecke: > @TymekDev and @PMassicotte can you please double check that Air works in various Quarto and RMarkdown documents? I have my doubts because I went looking in conform and could not find any specific Quarto support for LSP/formatting passthrough.
> 
> If it does seem to be working, I'd love to know what I missed so I can study how it works on the conform side

@DavisVaughan conform.nvim supports injected language formatting this probably what is being meant here, the see the following: https://github.com/stevearc/conform.nvim/blob/master/doc/advanced_topics.md#injected-language-formatting-code-blocks

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,52 @@
+---
+title: "Neovim"
+editor:
+  markdown:
+    wrap: sentence
+    canonical: true
+---
+
+# Installing the air client
```
lionel-, line 9:
```suggestion
# Installing the Air server
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,52 @@
...
+
+First, install the Air client by following the instructions on the [Air CLI documentation](https://posit-dev.github.io/air/cli.html).
```
lionel-, line 11:
```suggestion
First, install Air by following the instructions on the [Air CLI documentation](https://posit-dev.github.io/air/cli.html). The CLI binary bundles the Air server as well as a command line formatting tool.
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,52 @@
...
+
+# Configuration
+
+Air can be configured as a Language Server Protocol (LSP) client via [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) or as a formatter plugin via [conform.nvim](https://github.com/stevearc/conform.nvim).
```
lionel-, line 15:
```suggestion
Air can be configured as a Language Server Protocol (LSP) server via [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) or as a formatter plugin via [conform.nvim](https://github.com/stevearc/conform.nvim).
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,52 @@
...
+
+## LSP Configuration with nvim-lspconfig
+
+To configure Air's language server, add the following to your config:
```
lionel-, line 19:
```suggestion
To configure the client side to use Air's language server, add the following to your config:
```

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,52 @@
...
+
+```lua
+require("lspconfig").air.setup({})
+```
+
```
lionel-, line 24:
```suggestion

This is the recommended way to use Air. While the language server currently only supports formatting, it will gain more features in future.

```

lionel-: Any thoughts about layering Air as an LSP with `languageserver`?

TymekDev: @lionel- do you mean having both Air and `languageserver` enabled, but using only Air for formatting?

docs/editor-neovim.qmd
TymekDev: When responding to another comment, I just realized that `on_attach` seems to make the all-in-one `setup()` call:
```lua
require("lspconfig").air.setup({
  on_attach = function(_, bufnr)
    vim.api.nvim_create_autocmd("BufWritePre", {
      buffer = bufnr,
      callback = function()
        vim.lsp.buf.format()
      end,
    })
  end,
})
```

docs/editor-neovim.qmd
PMassicotte: Want to add it to the config instruction?

PMassicotte: Maybe something like

If both `Air` and `languageserver` are installed, you can use the following configuration to rely solely on `Air`'s formatting capabilities:

```lua
require("lspconfig").r_language_server.setup({
    on_attach = function(client, _)
        client.server_capabilities.documentFormattingProvider = false
        client.server_capabilities.documentRangeFormattingProvider = false
    end,
})
```

TymekDev: I'd say it's to disable `languageserver` formatting. In theory, someone can have other LSPs attached. Either that or using the `filter` in `vim.lsp.buf.format()` that you showed.

I think the bottom line is: make sure the user is aware they need to change something as both of these have formatting capabilities.

docs/editor-neovim.qmd
PMassicotte: I will use your solution, it makes sens and likely easier to configure for neovim newcomer. Thank you.

PMassicotte:
I think I have managed to include everyone's comment. Thank you very much.

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,59 @@
...
+                vim.lsp.buf.format()
+            end,
+        })
+    end,
+})
+```
+
+This is the recommended way to use Air. While the language server currently only supports formatting, it will gain more features in future.
+
+If both `Air` and `languageserver` are installed, you can use the following configuration to disable languageserver formatting, ensuring that only Air handles formatting:
```
lionel-, line 36: Perfect thanks!

lionel- (approved):

lionel-:
One question, is it possible to use Air as a language server in general and still configure conform to use it for Quarto documents at the same time?

lionel-:
I think this bit from their README indicates that both can be used simultaneously:

> Fixes bad-behaving LSP formatters - Some LSP servers are lazy and simply replace the entire buffer, leading to the problems mentioned above. Conform hooks into the LSP handler and turns these responses into proper piecewise changes.

PMassicotte:
> One question, is it possible to use Air as a language server in general and still configure conform to use it for Quarto documents at the same time?

Yes, it can be used for both, R files with the lsp-config and quarto/rmd when adding the conform configuration.

docs/editor-neovim.qmd
```diff
@@ -0,0 +1,59 @@
...
+require("lspconfig").air.setup({
+    on_attach = function(_, bufnr)
+        vim.api.nvim_create_autocmd("BufWritePre", {
+            buffer = bufnr,
+            callback = function()
+                vim.lsp.buf.format()
+            end,
+        })
+    end,
+})
```
DavisVaughan, line 31: Can someone help me understand why the `on_attach` is required?

I'm comparing to ruff, for example:
https://docs.astral.sh/ruff/editors/setup/#neovim

That does not have this `on_attach` hook to enable the `BufWritePre` pre-write hook. I would have thought that just setting up Air as an LSP would have activated something like this on its own.

Is Ruff's documentation missing something? Is this `BufWritePre` overkill? Or am I missing something else?

TymekDev: Technically, it's not reuiqred. However, without it formatting will work only on demand (by manually running `vim.lsp.buf.format()`). Adding this particular `on_attach` effectively enables formatting on save.

DavisVaughan: Ok so IIUC its like `editor.formatOnSave` in VS Code. Without that, you can invoke `Format Document` to tell the LSP to format your document, but with that setting set to `true` it will do so automatically on every save.

I guess in ruff they just didn't think to say that part.

But makes sense to me thanks!

DavisVaughan:
Thanks @PMassicotte, @TymekDev, and @m-muecke!

PMassicotte:
> Thanks @PMassicotte, @TymekDev, and @m-muecke!

Thank you everyone!

Issue #285, opened 2025-03-15
Review incorrect VS Code extension activation events


DavisVaughan:
With my newly acquired knowledge of glob syntax, I can say that `workspaceContains:*.R` and `workspaceContains:*.r` represent "any R file at the workspace root". This is not what we were going for, we wanted Air to be started if there were any R files in any directory under the workspace root. That would have been `workspaceContains:**/*.R`. You can confirm that Air doesn't immediately start up by opening a project with R files in a subfolder (and no editors open) and looking for `Air Language Server` in the Output pane. It won't be there.

I'm not actually sure we should use `**/*.R` though. This comment notes that it can be expensive in large projects https://github.com/sourcery-ai/sourcery/issues/112 (that comment also confirms my beliefs about how this glob syntax works).

In monorepos like Arrow, that could be _super_ slow, and they also probably don't want Air active when they are working on, say, C++ things.

Instead I think we should be like the vscode-python extension (https://github.com/microsoft/vscode-python/blob/6b7d8d1c12189126dce4c5d6733a74ab7762fb00/package.json#L71) and look for certain files at the workspace root that would be a good signal to preemptively activate on. I'm thinking:
- `air.toml`
- `.air.toml`
- `DESCRIPTION` (good for packages and projects that use this file)
- maybe something renv related? 

We would still keep `"onLanguage:r"` so that if any R file is opened the extension boots up, but this would allow it to preemptively start up in R packages and in any repo with `air.toml` at the workspace root.

lionel-:
I think this all makes good sense but we'd want to keep the `*.R` activation for simple non-package projects.

PR #286, opened 2025-03-15
Refine activation events

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/285

I didn't end up adding anything renv related, but I did add `*.Rproj` with the assumption that if you are opening a project that you previously used to open in RStudio, then you probably want Air to activate.

### DIFF
--- a/editors/code/CHANGELOG.md
+++ b/editors/code/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## Development version
 
+- The extension now activates automatically when an `air.toml`, `DESCRIPTION`, or `.Rproj` file is detected at the workspace root. Previously, the extension only activated after an R file was opened (#285).
+
 
 ## 0.8.0
 
--- a/editors/code/CHANGELOG.md
+++ b/editors/code/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ## Development version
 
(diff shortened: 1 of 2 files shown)


editors/code/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 ## Development version
 
+- The extension now activates automatically when an `air.toml`, `DESCRIPTION`, or `.Rproj` file is detected at the workspace root. Previously, the extension only activated after an R file was opened (#285).
```
DavisVaughan, line 10: (Technically it also used to activate if any `*.R` file was detected at the workspace root, but I imagine that was rare for our current set of users and is not worth mentioning)

editors/code/package.json
```diff
@@ -27,8 +27,10 @@
 	"icon": "air.png",
 	"activationEvents": [
 		"onLanguage:r",
-		"workspaceContains:*.r",
-		"workspaceContains:*.R"
```
lionel-, line 31: I think we should keep these for data analysis projects.

DavisVaughan:
We did end up adding top level `*.R` and `*.r` detection back.

We recognize this does not recursively check for R files to try and activate the extension early, and we think this is a good thing (both for file search performance and for accidentally activating the extension too early in a monorepo).

But we do think that if you have some R files at top level, there is probably a good chance you want Air support, so we can use it as one of our heuristics.

Starting up the Air extension is very fast right now, so all of this is just a bit of preoptimization. But it may come into play more if we do more on startup (rust-analyzer is a good example).

PR #287, opened 2025-03-20
docs: add link to GitHub repo?

maelle:
This would link to the docs source which might be more correct than just the repo. My initial motivation was to easily find Air source from the docs. :slightly_smiling_face: 

https://quarto.org/docs/websites/website-navigation.html#github-links


DavisVaughan:
Thanks!

PR #303, opened 2025-04-10
quarto & rmd need "injected" for conform.nvim/air formatting to work

idavydov:
I believe the minimal configuration in the documentation with nvim won't actually work unless `"injected"` is used. At least it didn't work on my setup.

@PMassicotte what do you think?

### DIFF
--- a/docs/editor-neovim.qmd
+++ b/docs/editor-neovim.qmd
@@ -62,8 +62,8 @@ Conform can be configured by adding the following to your `nvim/lua/plugins/conf
 ``` lua
 require("conform").setup({
     formatters_by_ft = {
-        quarto = { "injected", "prettier" },
-        rmd = { "injected", "prettier" },
+        quarto = { "injected" },
+        rmd = { "injected" },
         r = { "air" },
     },
 })
--- a/docs/editor-neovim.qmd
+++ b/docs/editor-neovim.qmd
@@ -62,6 +62,8 @@ Conform can be configured by adding the following to your `nvim/lua/plugins/conf
 ``` lua
 require("conform").setup({
     formatters_by_ft = {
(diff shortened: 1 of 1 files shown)


PMassicotte:
It is not needed. This is only used to format the markdown content of the document.

idavydov:
I added commits to only keep `"injected"` and remove `"prettier"`.

PMassicotte:
Here is what I have in my `conform.lua`

```lua
formatters_by_ft = {
    quarto = { "injected", "prettier" },
    r = { "air" },
    rmd = { "injected", "prettier" },
}

conform.formatters.prettier = {
    options = {
        ft_parsers = {
            quarto = "markdown",
            rmd = "markdown",
        },
    },
}
```
We could add that in the documentation.

PMassicotte:
My guess is tha when conform sees `quarto = { "injected", "prettier" },` it will use the chunk language to determine which formatter to use.

PMassicotte:
I have the same understanding. And you are right,

```lua
formatters_by_ft = {
    quarto = { "injected" },
    r = { "air" },
}
```

is enough to make it work (just tested it).

PMassicotte:
> None of ^ seems to actually handle the R code blocks themselves, so then the `"injected"` part of `quarto = { "injected", "prettier" }` says "parse a quarto doc using its formatter (prettier?? markdown?? hardcoded tree-sitter-markdown??), extract out the code blocks, and format each code block according to its own language specific formatter"

Exactly my understanding.

idavydov:
> is enough to make it work (just tested it).

This is what's currently in my branch (+ same for `rmd`); if the PR is reopened I think it will synchronize probably.

DavisVaughan:
Thanks all!

Issue #274, opened 2025-03-03
Explore `Air: Format project` command for VS Code for integration of Air into existing projects


DavisVaughan:
See https://github.com/r-lib/usethis/pull/2109#issuecomment-2695466329

The current initial workflow for Air in Positron for an _existing_ project looks like:
- Install Air CLI
- Install Air OpenVSX Extension
- `usethis::use_air()`
- Commit
- `air format .` one time using the Air CLI to get the whole project in order
- Commit
- Switch to Format on Save using the Air Extension from here on out

The current initial workflow for Air in RStudio for an _existing_ project looks like:
- Install Air CLI
- `usethis::use_air()`
- Commit
- `air format .` one time using the Air CLI to get the whole project in order
- Commit
- Switch to Format on Save using Air as an "external formatter" from here on out

I don't think we can improve on RStudio any further unless it bundles Air. But with VS Code / Positron I'd like to remove the CLI from the picture so it looks like:
- Install Air OpenVSX Extension
- `usethis::use_air()`
- Commit
- Run `Air: Format project` using the Air Extension
- Commit
- Switch to Format on Save using the Air Extension from here on out

I don't think having `use_air()` try and invoke Air will be super helpful here, as it will have a pre-req that Air is installed as a command line tool, and in our ideal world that actually won't be the case for Positron users (it'll be bundled in the extension but not available at the command line unless they install it themselves)

---

I think `Air: Format project` will be a little tricky because it is not natively part of the LSP and probably never will be. I think there are some issues in upstream VS Code requesting this feature that have been there for years. A few thoughts on it:

- It will likely be VS Code and Positron specific
- It probably will not go through the LSP? We essentially would have to perform `did_open`, `format`, `did_close` on every single discovered document that we need to format.
- We probably want to make sure every R editor is saved and closed before performing the format.
    - Ensuring the editors are closed should help the LSP stay in sync after the project wide formatting occurs. It'll just get a new did-open notification next time the user opens an R file with the refreshed contents.
- We probably want to refuse to format if there are any unstaged Git changes
- We can likely move `air format .` infra to a more generic place and reuse bits of that (but see adjacent idea below)
- It would likely require an `air.toml` to exist in the project so that we don't have to worry about forward propagation of editor specific settings (the `air.toml`s presence overrides forward propagation). That makes it nicely aligned with the `air format .` behavior.
- I'm not sure how we'd report errors about failures in individual files.

Possible adjacent idea:
[...]

PR #310, opened 2025-04-17
WIP: Nothing to see here

DavisVaughan:
Throwing everything up so I can more easily see it all together

Issue #110, opened 2024-12-16
Add hierarchical `air.toml` support


DavisVaughan:
https://docs.astral.sh/ruff/configuration/#config-file-discovery

> Unlike [ESLint](https://eslint.org/docs/latest/use/configure/configuration-files#cascading-configuration-objects), Ruff does not merge settings across configuration files; instead, the "closest" configuration file is used, and any parent configuration files are ignored

DavisVaughan:
Add nested `air.toml`s for https://github.com/posit-dev/air/pull/218 after we solve this

PR #312, opened 2025-04-18
Implement `Format Workspace Folder` command 

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/274

This PR implements a new VS Code / Positron specific command - `Format Workspace Folder`

The goal of this feature is to allow existing projects to adopt Air without having to leave the IDE or download a separate binary. The upcoming way to adopt Air in a project in Positron (where the extension will be already installed!!) looks like:
- `usethis::use_air()`
- `Air: Format Workspace Folder`
- Commit and push

Compare that with what we've been recommending:
- **Terminal:** Install Air CLI (ugh)
- Install Air OpenVSX Extension
- `usethis::use_air()`
- **Terminal:** `air format .` one time using the Air CLI to get the whole project in order
- Commit and push

Removing the terminal from the equation is a big win IMO. And removing the unnecessary separate binary install is also a big win, especially for first adoption where it should be insanely simple to get going.

https://github.com/user-attachments/assets/a09b07fc-34f1-4b32-b0cb-fd9be884ee44

## A CLI based approach

We've landed on using the embedded Air CLI as the engine for this feature. Notably we must first save and close all editors that live under the chosen workspace folder before calling the CLI, which ensures that any "dirty" editor contents are written to disk before calling the CLI. Ideally this means that the CLI, the LSP client, and the Air LSP server all stay in sync and use the on-disk contents as their "source of truth".

## Alternative 1 - UI open / format / save / close

There are extensions out there that "format a whole directory" for you. But those:
- Are extremely slow with large projects
- Will format non-R related files as well

Ultimately they just use vs code typescript apis to do:

```
for file in directory {
  openFile(file)
  formatFile(file)
  saveFile(file)
  closeFile(file)
}
```

This results in a very bad user experience IMO.

## Alternative 2 - Why not use a `WorkspaceEdit`?

While the LSP does not have official support for "Format a whole workspace folder", there is a concept of a multi-file `WorkspaceEdit`. Think of it as being able to supply `Vec<(Uri, TextEdit[])>` all at once as a bulk change.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit

I explored that in great depth in this PR https://github.com/posit-dev/air/pull/310, specifically in:

https://github.com/posit-dev/air/compare/535b5bfe00adbd4f3142b421875a07fd8dcf849a..8cd4a69d5b007aaa13063f0fc980e2c6a4639069

This does have the benefit of being fully in spec, but the actual implementation of it made me too nervous:

- It requires setting the metadata `isRefactoring: true` to avoid _opened_ all unsaved changed files for the user to manually review and save and close. While `WorkspaceEdit` is old, `WorkspaceEditMetadata` is only a [proposed feature of 3.18](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#workspaceEditMetadata) right now. I worked around this by calling the VS Code extension API of `vscode.workspace.applyEdit()` directly, which is the extension API equivalent of the LSP command, but it's still annoying.

- Even with `isRefactoring: true`, if you change exactly 1 file it will still open in a "dirty" state due to https://github.com/microsoft/vscode/issues/246720, meh.

- I think `WorkspaceEdit` is intended for large scale _refactoring_ but not necessarily _formatting_. Unfortunately when it applies the edit it performs a save action that _triggers formatting a second time_, because it has no way to know the edit itself was already a formatting edit. https://github.com/microsoft/vscode/issues/246720#issuecomment-2813296884

- It's just slow. Running it on `r-svn` which changes ~800 files at once will sometimes cause the editor to hang entirely. I couldn't figure out if it was because our diff algorithm was slow (which the CLI doesn't use at all), or if VS Code was having trouble processing such a huge batch of edits. It's entirely possible this is Air's fault, but it was enough to put me off the idea.

### DIFF
--- a/editors/code/src/command/workspace-folder-formatting.ts
+++ b/editors/code/src/command/workspace-folder-formatting.ts
@@ -13,22 +13,18 @@ import { isError, isResult, runCommand } from "../process";
  * - If 1 workspace folder is open, automatically uses it
  * - If >1 workspace folders are open, asks the user to choose
  *
- * # Tab closing
+ * # Tab saving
  *
  * Because we use the embedded Air CLI to perform the formatting, we force all
- * relevant tabs to be saved and closed before proceeding. This has two main
- * benefits:
+ * relevant tabs to be saved before proceeding. The Air CLI can't know about any
+ * in-memory changes that may have occurred in "dirty" editors. Forcing them to
+ * save causes any changes to be written to disk first, syncing VS Code with the
+ * CLI.
  *
- * - The Air CLI can't know about any in-memory changes that may have occurred
- *   in "dirty" editors. Forcing them to save and close causes any changes
- *   to be written to disk first, syncing VS Code with the CLI.
- *
- * - The LSP server part of Air needs to perfectly track changes for any open
- *   files. Closing the tabs first switches the server's "source of truth" from
- *   the client version of the file to the on-disk version of the file, which
- *   should ensure the LSP server won't have any chance of getting out of sync.
- *   This may be overkill, since the LSP server seems to handle git branch
- *   changes well, but it's still not a bad idea.
+ * We don't force the tabs to be closed. We've seen that VS Code seems to sync
+ * the LSP client and server well when performing a git branch switch (as long
+ * as all files are saved beforehand), so we are hopeful it should also behave
+ * well when we format with the CLI.
  *
  * # Forward propagation of settings
  *
@@ -55,8 +51,9 @@ export function workspaceFolderFormatting(ctx: Ctx): Cmd {
 			return;
 		}
 
-		const allTabsClosed = await closeAllTabs(workspaceFolder);
-		if (!allTabsClosed) {
+		const allSaved =
+			await saveAllDirtyWorkspaceTextDocuments(workspaceFolder);
+		if (!allSaved) {
 			return;
 		}
 
@@ -174,38 +171,24 @@ async function selectWorkspaceFolderFromQuickPick(
 }
 
 /**
- * Close all open editor tabs relevant to the workspace folder
+ * Save all open dirty editor tabs relevant to the workspace folder
  *
  * - Filters to only tabs living under the chosen workspace folder
- * - Asks the user if they are okay with us closing the editor tabs
- * - Asks the user to save or discard any dirty editor tabs
+ * - Asks the user if they are okay with us saving the editor tabs
  */
-async function closeAllTabs(
+async function saveAllDirtyWorkspaceTextDocuments(
 	workspaceFolder: vscode.WorkspaceFolder,
 ): Promise<boolean> {
-	// Collect all tabs from all tab groups
-	const allTabs = vscode.window.tabGroups.all.flatMap((group) => group.tabs);
+	const textDocuments = dirtyWorkspaceTextDocuments(workspaceFolder);
 
-	// Filter down to only tabs containing a text based resource who's uri
-	// prefix matches the workspace folder we are going to format. This way
-	// we don't have to close anything outside the workspace folder.
-	const allRelevantTabs = allTabs.filter((tab) => {
-		const input = tab.input;
-
-		if (!(input instanceof vscode.TabInputText)) {
-			return false;
-		}
-
-		return input.uri.fsPath.startsWith(workspaceFolder.uri.fsPath);
-	});
-
-	if (allRelevantTabs.length === 0) {
-		// Nothing to close!
+	if (textDocuments.length === 0) {
+		// Nothing to save!
 		return true;
 	}
 
+	// Ask the user if we can save them
 	const answer = await vscode.window.showInformationMessage(
-		`All editors within the ${workspaceFolder.name} workspace folder must be closed before formatting. Proceed with closing these editors?`,
+		`All editors within the ${workspaceFolder.name} workspace folder must be saved before formatting. Proceed with saving these editors?`,
 		{ modal: true },
 		"Yes",
 		"No",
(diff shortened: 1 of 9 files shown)


editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+ * exist, those settings override document settings.
+ *
+ * Because we go through the Air CLI here, forward propagated document settings
+ * are ignored, so even if you don't have an `air.toml`, you will get Air's
+ * default indentation settings.
+ *
+ * We can't simply require an `air.toml` at the workspace folder root to avoid
+ * confusion here, because you may instead have an `air.toml` at
+ * `{root}/R/air.toml` that would handle any R files in your project. Instead,
+ * we hope this isn't common enough to come up much in practice.
```
DavisVaughan, line 47: Notable point of potential confusion, but hopefully not too bad.

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+			const answer = await vscode.window.showInformationMessage(
+				`Errors occurred while formatting the ${workspaceFolder.name} workspace folder. View the logs?`,
+				{ modal: true },
+				"Yes",
+				"No",
+			);
+
+			if (answer === "Yes") {
+				output.show();
+			}
```
DavisVaughan, line 113: I didn't show it, but if any files fail to format we write the emitted stderr into the logs and offer to take the user there

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	if (workspaceFolders.length === 1) {
+		return workspaceFolders[0];
+	}
+
+	// Let the user select a workspace folder if >1 are open, may be
+	// `undefined` if user bails from quick pick!
+	const workspaceFolder =
+		await selectWorkspaceFolderFromQuickPick(workspaceFolders);
```
DavisVaughan, line 143: In the common case of just 1 workspace folder, we don't even ask the user which to use

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * Because we use the embedded Air CLI to perform the formatting, we force all
+ * relevant tabs to be saved and closed before proceeding. This has two main
+ * benefits:
+ *
+ * - The Air CLI can't know about any in-memory changes that may have occurred
+ *   in "dirty" editors. Forcing them to save and close causes any changes
+ *   to be written to disk first, syncing VS Code with the CLI.
+ *
+ * - The LSP server part of Air needs to perfectly track changes for any open
```
lionel-, line 26: I'm still not sure about that. I think I'd prefer to keep them open for now as it gives a better UX, and only close them if it turns out to be needed. As you noted the IDE doesn't generally need to close files when the whole repo is changed from under its feet via git, so I wouldn't expect problems here.

editors/code/src/command/workspace-folder-formatting.ts
lionel-: I was thinking this command should ensure an air.toml file at the project root. Letting the cli potentially pick up config files up the folder hierarchy seems unexpected to me.

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+		// i.e., `air format {workspaceFolderPath} --no-color`
+		const args = ["format", workspaceFolderPath, "--no-color"];
+
+		// This should not matter since the path is explicitly supplied, but better to be safe
+		const options = {
+			cwd: workspaceFolderPath,
+		};
+
+		// A promise that resolves when the spawned process closes or errors
+		const finishedFormatting = new Promise<void>((resolve) => {
```
lionel-, line 77: Just noting that this should be extracted in a helper if we call a binary again (I'm slightly surprised there is no higher level API for this).

editors/code/src/command/workspace-folder-formatting.ts
lionel-: Nice!

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -0,0 +1,228 @@
...
+		"No",
+	);
+
+	if (answer !== "Yes") {
+		// User said `"No"` or bailed from the menu
+		return false;
+	}
+
+	// Close all tabs at once, each dirty tab will prompt the user to save or discard any changes
+	return await vscode.window.tabGroups.close(allRelevantTabs);
```
lionel-, line 227: oh I thought we'd ask the user if it's okay to save files and do it all at once silently if they say yes.

editors/code/src/command/workspace-folder-formatting.ts
DavisVaughan: The comment there explains why it can't just ensure an air.toml is at the project root

DavisVaughan: We currently look _up_ the file hierarchy for `air.toml` but both the CLI and LSP are also going to also look _down_ the file hierarchy for them as well. https://github.com/posit-dev/air/issues/110 When that happens, it will be perfectly reasonable to open, say, the top level `arrow/` workspace to work on the `arrow/r` package nested inside it, and the `air.toml` would be at `arrow/r/air.toml`, i.e. "below" the workspace

DavisVaughan: > Letting the cli potentially pick up config files up the folder hierarchy seems unexpected to me.

Also, both the LSP and the CLI look _up_ the folder hierarchy in the same way. If you open a workspace at `dplyr/R` and your `air.toml` was at `dplyr/`, even though that is "outside" the LSP workspace, we still find and use it. I think that is the only way to have a consistent experience across LSP and CLI.

editors/code/src/command/workspace-folder-formatting.ts
DavisVaughan: I looked and ended up copying what Go does

editors/code/src/command/workspace-folder-formatting.ts
DavisVaughan: Okay, we now save dirty open text documents that live under the workspace root, but we no longer close them

This does feel like it aligns with how a git branch switch works, so 🤞

editors/code/src/command/workspace-folder-formatting.ts
DavisVaughan: Done!

editors/code/src/command/workspace-folder-formatting.ts
lionel-: > Running Format Workspace Folder from the arrow/ workspace should be allowed without requiring an air.toml at arrow/air.toml, which wouldn't make sense.

Why not open the `arrow/r` folder and perform the command? That would make more sense to me than do it at top level and perhaps format files in other subfolders using an air.toml file lying around in a parent folder.

Or if we think that's not good UX, just add `Format Workspace Subfolder` and let user quickpick a folder in the project, and that would add the air.toml file at that place.

DavisVaughan: We've decided that as long as we push most people towards running `Air: Setup Workspace Folder` (https://github.com/posit-dev/air/issues/323) first (through documentation), then this will mostly be a non-issue, because an `air.toml` will exist through the setup process.

That allows rare monorepo people like `arrow/` devs to skip that setup step, instead adding configuration to `arrow/r/air.toml`, and still be able to call `Air: Format Workspace Folder` from `arrow/` without being forced into a top level `air.toml`.

Issue #268, opened 2025-02-27
Official Air GitHub Actions for formatting


DavisVaughan:
The Shiny one has a lot of the customization I'd want
https://github.com/rstudio/shiny-workflows/blob/main/format-r-code/action.yaml

- Pinning a version
- Formatting followed by committing if any files were formatted
- Check mode that errors rather than committing
- Default formats `.` but provide some way to customize the path(s), or maybe a way to customize the working directory

I'm not sure if we want it to be a composite workflow like the Shiny one or if we want to, say, always make it run on `ubuntu-latest`. 

This probably needs to be its own repo so we can use tags to version the workflow

The ruff one is fairly sophisticated, and is not a composite workflow
https://github.com/astral-sh/ruff-action/tree/main

The ruff one even lives in the official marketplace
https://github.com/marketplace/actions/ruff-action

PR #324, opened 2025-04-25
Add support for `--check=github` as an output format

DavisVaughan:
Part of #268 
Goes with https://github.com/posit-dev/air-action

I'm hoping that our example action will be able to set `air format . --check=github` to get GitHub approved annotations in PRs.

You can now do:

```bash
# Same as before
air format . --check 

# Explicitly request the above
air format . --check=interactive

# GitHub flavored file level annotations
air format . --check=github
```

I'm thinking in the future we could also have `air format . --check=diff` to generate diff output

---

I also got very tired of mostly "flying blind" with our CLI output, so I added a way for us to call the actual `air` cli binary from our tests.

It turns out that for integration tests, Rust already goes ahead and compiles the binary for this exact purpose, and at compile time it exposes `CARGO_BIN_EXE_air` as an env var which holds the path to the binary. So I've added some utilities that use that and `Command::new()` to run air commands and capture their stderr / stdout. I wish they were unit tests but 🤷


DavisVaughan:
Not doing this after all, not needed. Extracted test upgrades into #329

PR #329, opened 2025-04-29
Use integration tests in `air/` to directly test the `air` binary

DavisVaughan:
While we do currently have [some air cli related tests](https://github.com/posit-dev/air/blob/8668362a4c20a4e5e73fcb86c6404f9d86306d30/crates/air/src/commands/format.rs#L241-L377), they currently have a decent number of limitations:

- They aren't actually testing the `air` binary, so we have a gap in what we see vs what the user sees
- We can't test (i.e. snapshot) the stdout / stderr of the binary
- You can only call `run()` once per test, as that sets up Air's global logging and can only be done once per process

I've come to realize that the cli crate is the one place we actually do want to use integration tests because of this very neat feature of integration tests: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#integration-tests

> Binary targets are automatically built if there is an integration test. This allows an integration test to execute the binary to exercise and test its behavior. The CARGO_BIN_EXE_<name> [environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) is set when the integration test is built so that it can use the [env macro](https://doc.rust-lang.org/std/macro.env.html) to locate the executable.

In other words, during integration tests an actual `air` binary is built, and cargo totally intends for us to have access to it! So lets use that!

This gives us the VERY powerful ability to snapshot test the stdout / stderr of the air cli - which is something we've been flying blind on before now. This is going to be so nice!

I've set up the integration test directory structure following the advice of https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html for two reasons:
- One integration test binary is faster to compile
- When you have two integration test binaries and a shared `helpers/` folder between them, each integration test binary must use EVERY helper function, otherwise a helper function unused by at least 1 integration test binary gets marked as "dead" code to be removed by rustc. It's a known annoying issue, but this structure gets around it nicely https://github.com/rust-lang/rust/issues/46379


crates/air/src/commands/format.rs
```diff
@@ -98,7 +98,11 @@ impl FormatMode {
 
 fn inform_changed(paths: &[PathBuf], f: &mut impl Write) -> io::Result<()> {
     for path in paths.iter().sorted_unstable() {
-        writeln!(f, "Would reformat: {}", path.display())?;
+        writeln!(
+            f,
+            "Would reformat: {path}",
+            path = relativize_path(path).underline()
+        )?;
```
DavisVaughan, line 105: I made this one change to cli behavior so that snapshots are relative to the working directory and stable when run on CI.

In retrospect, it actually aligns with how we emit paths in error messages, so I think this is more correct anyways.

crates/air/tests/integration/format.rs
```diff
@@ -0,0 +1,205 @@
...
+    let path = "test.R";
+    std::fs::write(directory.join(path), "1 + 1")?;
+
+    let output = Command::new(binary_path())
+        .current_dir(directory)
+        .arg("format")
+        .arg(path)
+        .run();
+
+    assert!(output.status.success());
```
DavisVaughan, line 24: Here's where we call the `air` binary. `binary_path()` and `run()` are my own creations, but mostly it's a standard `Command` call, pretty cool!

We can either inspect the `output` manually, or `insta::assert_snapshot!()` it (as i do in `--check` tests below)

crates/air/tests/integration/format.rs
```diff
@@ -0,0 +1,205 @@
...
+    let path2 = path2.to_str().unwrap();
+
+    insta::assert_snapshot!(Command::new(binary_path())
+        .current_dir(path_root())
+        .arg("format")
+        .arg(path1)
+        .arg(path2)
+        .arg("--check")
+        .run());
+}
```
DavisVaughan, line 205: These are new tests that would have been impossible to create before now because we had no way to capture stderr/stdout

Such a big win in test ergonomics

crates/air/tests/integration/help.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[test]
+fn test_format_help() {
+    insta::assert_snapshot!(Command::new(binary_path()).arg("format").run());
+    insta::assert_snapshot!(Command::new(binary_path())
+        .arg("format")
+        .arg("--help")
+        .run());
+    insta::assert_snapshot!(Command::new(binary_path()).arg("format").arg("-h").run());
+}
```
DavisVaughan, line 22: When working on the now defunct https://github.com/posit-dev/air/pull/324, it would have been really nice to have snapshot tests for the help docs

crates/air/tests/integration/helpers/command_ext.rs
```diff
@@ -0,0 +1,73 @@
...
+        let arguments = self.arguments.join(" ");
+
+        // Normalize all `\` to `/` for snapshot stability
+        let stdout = stdout.replace('\\', "/");
+        let stderr = stderr.replace('\\', "/");
+        let arguments = arguments.replace('\\', "/");
+
+        // Replace Windows help documentation's `air.exe` with `air` for snapshot stability
+        let stdout = stdout.replace("air.exe", "air");
+        let stderr = stderr.replace("air.exe", "air");
```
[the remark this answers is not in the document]
DavisVaughan: Maybe `set_snapshot_path()` set to the OS name? But that is likely going to be frustrating for us to manage because we'd need a windows computer to update the snapshots on each time.
https://docs.rs/insta/latest/insta/struct.Settings.html#method.set_snapshot_path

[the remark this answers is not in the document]
DavisVaughan: Went with explicit as-needed normalization methods `normalize_os_path_separator()` and `normalize_os_executable_name()` that you see at the call site. Seems much better for long term debuggability

lionel- (approved):
Awesome!

Issue #277, opened 2025-03-04
The Positron settings explanation was maybe clearer in your announcement :-)


maelle:
I mean https://github.com/posit-dev/positron/discussions/6095 is IMHO clearer than https://posit-dev.github.io/air/editor-vscode.html because it indicates where to find the settings to change. Could the Air docs use that same text?

PR #343, opened 2025-05-11
Elaborate information on collaborative projects

xx02al:
I would find this additional information on collaborative projects helpful. Both, for people starting with air, as well as when referring your project colleagues to the air documentation.

Maybe closes [#277].


DavisVaughan:
Thanks!

PR #344, opened 2025-05-14
Autobrace for loops, while loops, repeat loops, and function definitions

DavisVaughan:
Branched from https://github.com/posit-dev/air/pull/340

Adds our remaining autobracing capabilities to round out the feature. These are significantly less complicated than if/else handling because we don't have to worry about nesting.

The loops all now use `FormatLoopBody`, and function definitions use `FormatFunctionBody`, which is similar in design to `FormatIfBody`.

It required quite a few comment tweaks, but the original implementation of comments for these nodes was incomplete, so that's expected. We now have quite a few extra tests for comment handling here, in addition to autobrace handling.


docs/formatter.qmd
```diff
@@ -174,6 +174,298 @@ list(foo, bar)
...
+``` r
+# Adding a line break before `is.logical(x)` forces autobracing
+result <- map_lgl(xs, function(x)
+    is.logical(x) && length(x) == 1L && !is.na(x))
+
+# Becomes:
+result <- map_lgl(xs, function(x) {
+  is.logical(x) && length(x) == 1L && !is.na(x)
+})
+```
```
DavisVaughan, line 427: I thought this was a cool section

lionel- (approved):

docs/formatter.qmd
```diff
@@ -174,6 +174,298 @@ list(foo, bar)
...
+fn <- function() {
+  if (condition) stop("oh no")
+  1 + 1
+}
+
+fn <- function() {
+  if (condition) return(1)
+  1 + 1
+}
+```
```
lionel-, line 261: I would show the resulting formatting and mention that having side effects like stop or return on their own line is important for readability as they impact control flow or state. You could also add a `<-` example.

docs/formatter.qmd
```diff
@@ -174,6 +174,298 @@ list(foo, bar)
...
+
+If you try and run this, then you'll see an error like `Error: unexpected 'else'`.
+This is particularly annoying when you're working inside a debugger.
+Most R debuggers allow you to pause inside functions and highlight and run chunks of that function.
+If you're paused inside `fn()` and try to highlight and run the if statement, then it will confusingly fail to parse.
+Autobracing multiline if statements avoids this problem entirely.
+
+## For, while, and repeat loops
+
+Air unconditionally autobraces the body of all R loops.
```
lionel-, line 316: Mostly for consistency with `if` constructs?

docs/formatter.qmd
lionel-: yep it's nice you can choose which format fits most for your use case

lionel- (approved):

PR #351, opened 2025-05-30
Never expand empty `{}` across multiple lines

### LINKED ISSUE #43: Probably allow empty `{}` to not expand after all
Image

Saw this in elmer. Would hate for the `{}` the be expanded here.

i.e. its sometimes useful to provide function "stubs" that just eat their input, but are required for some API interface contract

Possibly we do something similar with `tryCatch()`, not sure.

Will require special checking + more tests in `call_arguments.rs` as this will "enable" that most-flat case that currently is never hit

DavisVaughan:
Closes #43 

CC @hadley 

I've done what ya'll preferred and used the simple rule of "never expand empty `{}`", which was easy enough to implement.

All of the following are now allowed

```r
# the motivating case for this
dummy <- function() {}

# the "block until done" pattern
while (waiting()) {}

# seen in lintr discussions as an alternative to `NULL`
switch(x, a = {}, b = 2)

# seen in httr2
function(
  expr = {}
) {
  this_first()
  expr
}

# swallow error, return NULL
tryCatch(
  expr = {
      stop("oh no")
  },
  error = function(e) {}
)
```

We end up allowing this weirdness for if/else, but it's almost certainly a logic error, and we have decided in the past that we don't make formatter decisions based on logic errors

```r
if (condition) {} else {
  2
}

if (condition) {
  2
} else {}
```

### DIFF
--- a/crates/air_r_formatter/src/r/auxiliary/braced_expressions.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/braced_expressions.rs
@@ -1,3 +1,4 @@
+use crate::comments::RComments;
 use crate::prelude::*;
 use air_r_syntax::AnyRExpression;
 use air_r_syntax::RArgument;
@@ -7,7 +8,6 @@ use air_r_syntax::RExpressionList;
 use air_r_syntax::RSyntaxToken;
 use biome_formatter::format_args;
 use biome_formatter::write;
-use biome_formatter::CstFormatContext;
 use biome_rowan::SyntaxResult;
 
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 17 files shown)


crates/air_r_formatter/src/r/auxiliary/braced_expressions.rs
```diff
@@ -20,24 +20,61 @@ impl FormatNodeRule<RBracedExpressions> for FormatRBracedExpressions {
             r_curly_token,
         } = node.as_fields();
 
-        // Check if we are formatting empty braces, like `{ }`
-        if expressions.is_empty() {
-            return fmt_empty(&l_curly_token, &r_curly_token, node, f);
+        match braced_expressions_variant(node, f.comments()) {
```
DavisVaughan, line 23: Implementation wise the most important thing to look at is this nice categorization helper, `braced_expressions_variant()`.

We use it here and over in `call_arguments.rs`. Splitting `RBracedExpressions` into these 4 categories is _just_ the right amount of categorization to be maximally reusable

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -1239,11 +1221,25 @@ fn argument_is_groupable(argument: &AnyRExpression) -> SyntaxResult<bool> {
...
+fn braced_expressions_is_groupable(node: &RBracedExpressions, comments: &RComments) -> bool {
+    match braced_expressions_variant(node, comments) {
+        // These two stay collapsed as `{}` or `{{ x }}`, and don't benefit from grouping
+        BracedExpressionsVariant::Empty => false,
+        BracedExpressionsVariant::CurlyCurly(_) => false,
+        // These two expand across multiple lines, and benefit from grouping
+        BracedExpressionsVariant::EmptyWithDanglingComments => true,
+        BracedExpressionsVariant::NotEmpty => true,
+    }
+}
```
DavisVaughan, line 1245: Reuse of the `braced_expressions_variant()` helper to allow us to easily decide if we should consider using best fitting on this trailing argument or not

docs/formatter.qmd
```diff
@@ -489,6 +489,42 @@ if (condition) {
...
+switch(x, a = {}, b = 2)
+
+function(
+  expr = {}
+) {
+  this_first()
+  expr
+}
+```
+
```
DavisVaughan, line 527: A new small addition to the docs, seems worth calling this out

lionel- (approved):

PR #354, opened 2025-06-04
Use `biome_line_index` to avoid `biome_lsp_converters`

DavisVaughan:
Closes #353 in favor of this simpler approach (though there are a few things from that PR ill add in later on in another PR)

This is a minimal PR that gets rid of our dependency on `biome_lsp_converters` in favor of the lighter weight `biome_line_index`. We use `biome_lsp_converters` for two things:
- The LSP _agnostic_ `LineIndex` type, which is extremely useful for conversion between byte offset and line/column position
- A few LSP _specific_ conversion utilities:
    - `from_proto::offset()`
    - `from_proto::text_range()`
    - `to_proto::position()`
    - `to_proto::range()`

The painful thing about `biome_lsp_converters` is that it locks us in to `tower_lsp`, therefore also locking us to `tokio`, due to the fact that their LSP specific conversion utilities use `tower_lsp::lsp_types::*`.

I'm planning on sending a PR to biome (https://github.com/biomejs/biome/pull/6222) to request that they split out the LSP agnostic `LineIndex` type into a `biome_line_index` crate which I've gone ahead and copy/pasted in here. That allows us to use their fantastic `LineIndex` implementation, but critically we now _don't_ rely on tower-lsp through them in any way. I'm hopeful they accept that PR, at which point we can drop our shim in favor of their real thing (we will also need to sync with biome `main` to do that though, which may be painful)

This will allow us to switch over to lsp-server without any issues. Previously, we were blocked from doing this due to our deep reliance on tower-lsp through biome-lsp-converters.


crates/biome_line_index/src/lib.rs
DavisVaughan, line 1: Ignore, this is Biome code

crates/biome_line_index/src/line_index.rs
DavisVaughan, line 1: Ignore, this is Biome code

PR #364, opened 2025-06-13
Optimize `on_did_change()`

DavisVaughan:
Branched from #363 

It turns out that we are rebuilding our `LineIndex` _at least twice_ on every single `on_did_change()` call.
- One is here, where we ignore the fact that we already had a `LineIndex` built for this `Document` https://github.com/posit-dev/air/blob/be96e71284928f504a9402368bc716520fe15647/crates/lsp/src/rust_analyzer/utils.rs#L35
- One is here, after we've applied all changes https://github.com/posit-dev/air/blob/be96e71284928f504a9402368bc716520fe15647/crates/lsp/src/documents.rs#L135-L136

That's not super efficient! In the ~90% case where the user is just typing one character at a time, we should only be rebuilding the line index exactly 1 time.

I've folded `apply_document_changes()` into `on_did_change()` to keep all of the mutable-state-updating code close together.

The actual implementation is basically still the same, I've just had to rework things a little to do this optimization.

I had added in some logging and can confirm that we now only rebuild the `LineIndex` 1 time on every update as the user is typing. We typically also rebuild it exactly 1 time during a find-and-replace as well, due to how we take advantage of VS Code sending us the updates from bottom to top (so we can apply them sequentially without invalidating our line index).

I might try a PR upstream to rust-analyzer, because I'm fairly certain they have this too from looking at their code.

### DIFF
--- a/crates/lsp/src/documents.rs
+++ b/crates/lsp/src/documents.rs
@@ -5,8 +5,9 @@
 //
 //
 
-use settings::LineEnding;
 use std::ops::Range;
+
+use settings::LineEnding;
 use tower_lsp::lsp_types;
 
 use crate::line_index::LineIndex;
@@ -73,9 +74,7 @@ impl Document {
         // select `LineEndings::Auto`, and then pass that to `LineIndex`.
 
         // Create line index to keep track of newline offsets
-        let line_index = LineIndex {
-            index: triomphe::Arc::new(biome_line_index::LineIndex::new(&contents)),
-        };
+        let line_index = LineIndex::new(&contents);
 
         // Parse document immediately for now
         let parse = air_r_parser::parse(&contents, Default::default());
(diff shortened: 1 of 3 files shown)


crates/lsp/src/documents.rs
```diff
@@ -119,78 +125,69 @@ impl Document {
...
-        for change in content_changes {
-            // The None case can't happen as we have handled it above already
-            if let Some(range) = change.range {
-                if index_valid <= range.end.line {
-                    line_index = biome_line_index::LineIndex::new(&text);
-                }
-                index_valid = range.start.line;
-                if let Ok(range) = from_proto::text_range(range, &line_index, encoding) {
-                    text.replace_range(Range::<usize>::from(range), &change.text);
-                }
```
DavisVaughan, line 190: We used to just...skip (??) the file update if we were unable to convert from `Position` to `TextRange`.

This is a critical failure IMO and should be treated as a panic, which we now do

crates/lsp/src/documents.rs
```diff
@@ -119,78 +125,69 @@ impl Document {
...
-            }
-        }
-        text
+        // Rebuild the `line_index` after applying the final edit, and sync other fields
+        self.line_index = LineIndex::new(&self.contents);
```
DavisVaughan, line 188: For the cases of:
- A single full document change
- 1 or more incremental changes that don't invalidate each other

this is now the only place we recreate the line index

Hinkee (commented):
[](https://www.facebook.com/reel/586681813960375/?mibextid=rS40aB7S9Ucbxw6v)

Issue #302, opened 2025-04-09
Use custom cargo-dist binary building workflows


DavisVaughan:
In https://github.com/posit-dev/air/pull/301 we had to pin ourselves to Rust 1.83 due to the aarch64-pc-windows-msvc build and cargo-xwin.

cargo-xwin's docker contain pulls in 1.83 by default, and while that is configurable, it is not available to us to change. See the PR above for full info.

To free us to use Rust 1.84+, and to give us more control over the binary building process of cargo-dist in general, we should look into setting

```
# Whether CI should include auto-generated code to build local artifacts
build-local-artifacts = false
# Local artifacts jobs to run in CI
local-artifacts-jobs = ["./build-binaries"]
```

Where `build-binaries.yml` is our own workflow that builds binaries for all platforms of interest. For each target, we would be in charge of:
- Building the binary and wrapping it into a `tar.gz` (Unix) or `zip` (Windows)
- Computing the SHA
- Uploading both as GitHub artifacts

If we do both of those, cargo-dist should be able to pick up from there and generate the GitHub Release and the installer scripts, both of which are still very valuable parts of the cargo-dist process.

Ruff has done this. While they use maturin, I don't think we should let that deter us from looking at their approach:
https://github.com/astral-sh/ruff/blob/main/.github/workflows/build-binaries.yml

It's going to take some effort to get the CI set up correctly for all targets (cross compilation for ARM Windows, cargo-zigbuild usage or PyPa docker containers for old Linux support), but ultimately having more full over the build process is probably going to be good in the long run due to cargo-dist having no maintainer right now, and because the build process has been a source of frustration for us over the months.

DavisVaughan:
Note that we were able to update to Rust 1.87 https://github.com/posit-dev/air/pull/372 due to xwin updating its own Dockerfile to 1.87 as well

PR #372, opened 2025-06-20
Update to Rust 1.87 and Rust 2024 edition

DavisVaughan:
- [x] Resolve https://github.com/rust-cross/cargo-xwin/issues/159
- [x] Do a standard merge commit

Related to https://github.com/posit-dev/air/issues/302, which notes how we are tied to cargo-xwin's latest version in their Dockerfile for ARM Windows. Note that they have recently updated that Dockerfile to use Rust 1.87, so we should be able to use it too https://github.com/rust-cross/cargo-xwin/blob/main/Dockerfile.

Rust 2024 edition came with a few rustfmt updates. I ran `cargo fmt` on the whole project (note we don't use `+nightly` like we do in Ark). They look harmless.

I'll do a dry-run of building the binaries to ensure it works as expected on all platforms.


DavisVaughan:
I now see `RUST_VERSION=1.87.0` in my dry run in https://github.com/posit-dev/air/actions/runs/16141856360/job/45551023899#step:2:72 now that cargo-xwin has been updated and the docker image has been fixed

PR #376, opened 2025-07-06
Add brew formula option for macOS/Linux

coatless:
Update install options for macOS/Linux to include the Homebrew [air formula](https://formulae.brew.sh/formula/air) approach.


lionel-:
Thank you. For reference: https://github.com/Homebrew/homebrew-core/pull/228899

Can you tell us a bit more about the updating process? I guess it's not automated so the brew formula might lag behind?

PR #385, opened 2025-07-17
Add --output and --stdout options to format command

armish:
Allows formatting without overwriting the original file by either writing to a different file (`--output`) or to stdout (`--stdout`). Both options only work with single input files and are mutually exclusive with each other and `--check`.


armish:
Thank you for the explanation @DavisVaughan and apologies that I missed that issue. I was more focused on "don't-overwrite" behaviour and didn't see that this was already on your radar. Also no worries about dismissing the PR as it was such a trivial one that I thought it was easier to just send a PR.

PR #390, opened 2025-07-24
Implement `Air: Initialize Workspace Folder`

### LINKED ISSUE #323: `Air: Setup Workspace Folder`
Copy the VS Code parts of the `usethis::use_air()` function into a command, so setting up Air for the first time would be:
- `Air: Setup Workspace Folder`
- `Air: Format Workspace Folder`
 
Details:
- Requires a workspace folder to be open, asks user to select one if multiple are open (reuse `Format Workspace Folder` infra for this)
- Make empty `air.toml`
- Add or update `settings.json` 
- Add or update `extensions.json`

`settings.json`:

```json
{
    "[r]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "Posit.air-vscode"
    }
}
```

`extensions.json`

```json
{
    "recommendations": [
        "Posit.air-vscode"
    ]
}
```

DavisVaughan:
Closes #323 

This is a Positron and VS Code specific complement to `usethis::use_air()`, this way our setup recommendation for Positron doesn't need to involve R at all, it's just:
- `Air: Initialize Workspace Folder`
- `Air: Format Workspace Folder` (or click the button during initialization)

A few quirks:
- There is an API to update `settings.json`, but no API to update `extensions.json`. The `extensions.json` file is less critical, so I've taken an all or nothing approach where we just don't even try to do anything if the user already has one. I think it would even be more complex than I'd want to try and parse the existing one and see if we are recommended or not, and if not, prompt the user with what they could add. That all just doesn't seem worth it (and toast notifications aren't a great UI for that kind of prompt IMO).
- The `settings.json` update UI is a bit wonky for this. There isn't a great API combination for "I know the workspace folder, and now I'm trying to tweak _just one_ language specific setting". Instead you have to replace _all_ of `[r]` in bulk. You'll see some comments in there about how we have to do this.

https://github.com/user-attachments/assets/d48b733d-adf2-4243-b523-a9de519f27b9

### DIFF
--- a/editors/code/src/workspace.ts
+++ b/editors/code/src/workspace.ts
@@ -131,6 +131,7 @@ async function selectWorkspaceFolderFromQuickPick(
  * Save all open dirty editor tabs relevant to the workspace folder
  *
  * - Filters to only tabs living under the chosen workspace folder
+ *   (note that this rules out untitled editors)
  * - Asks the user if they are okay with us saving the editor tabs
  */
 export async function saveAllDirtyWorkspaceTextDocuments(


editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -42,195 +40,78 @@ import { isError, isResult, runCommand } from "../process";
  * `{root}/R/air.toml` that would handle any R files in your project. Instead,
  * we hope this isn't common enough to come up much in practice.
  */
-export function workspaceFolderFormatting(ctx: Ctx): Cmd {
-	return async () => {
```
DavisVaughan, line 46: There are no actual changes to this function, I've just de-dented it one level since it no longer returns a callback

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,206 @@
...
+	workspaceFolder: vscode.WorkspaceFolder,
+): Promise<void> {
+	const vscodeUri = vscode.Uri.joinPath(workspaceFolder.uri, ".vscode");
+	const extensionsUri = vscode.Uri.joinPath(vscodeUri, "extensions.json");
+
+	if (await fileExists(extensionsUri)) {
+		// Just bail if the user already has `extensions.json`, we don't
+		// try to update an existing one
+		return;
+	}
```
DavisVaughan, line 96: This is the most effort I think I'm willing to put in for now 😛

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,206 @@
...
+	const configR = config.inspect("[r]")?.workspaceFolderValue || {};
+	await config.update(
+		"[r]",
+		{
+			...configR,
+			"editor.formatOnSave": true,
+			"editor.defaultFormatter": "Posit.air-vscode",
+		},
+		vscode.ConfigurationTarget.WorkspaceFolder,
+	);
```
DavisVaughan, line 152: Wonky config updating scheme, I did check and other people also do it like this

For example, the eslint extension does it this way, and that's maintained by the main LSP guy
https://github.com/microsoft/vscode-eslint/blob/001edda5bd168a9c6a79cba767111ba21eadd052/client/src/settings.ts#L330

editors/code/src/command/workspace-folder-formatting.ts
```diff
@@ -42,195 +40,78 @@ import { isError, isResult, runCommand } from "../process";
...
-		},
-	);
-
-	if (!workspaceFolderName) {
-		// User bailed from the quick pick
-		return undefined;
-	}
+	if (anyErrors) {
+		const answer = await vscode.window.showInformationMessage(
+			`Errors occurred while formatting the ${workspaceFolder.name} workspace folder. View the logs?`,
```
lionel-, line 87: Nice

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,209 @@
...
+
+	await createAirToml(workspaceFolder);
+	await createExtensionsJson(workspaceFolder);
+	await updateSettingsJson(workspaceFolder);
+	await updateRbuildignore(workspaceFolder);
+
+	const message = `Successfully initialized the ${workspaceFolder.name} workspace folder.`;
+	const shouldFormatItem = `Click to format the ${workspaceFolder.name} workspace folder`;
+
+	const item = await vscode.window.showInformationMessage(
```
lionel-, line 39: So this resolves only when the toast notif is discarded one way or another?

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,209 @@
...
+ * - Doesn't destroy existing `[r]` or `[quarto]` settings that we aren't
+ *   updating (like `config.update("[r]", value)` naively does)
+ * - Doesn't pull extra inherited global settings (like `config.get()` would do)
+ *
+ * The trick is to `inspect()` all of the `[r]` configuration specific to just
+ * the `workspaceFolder` and bulk update it, retaining all old settings but
+ * overriding the ones we care about updating.
+ *
+ * This does unfortunately drop comments in the `[r]` and `[quarto]` sections,
+ * but we can't do better.
```
lionel-, line 130: I'm assuming that modifying the json via the workspace settings UI would also drop the comments? So it's more like settings comments are not supported in vscode/positron at all.

lionel-: oh hmm, just tried it and they are preserved.

lionel-: Looks like the editing UI uses `jsonc-parser` (https://github.com/microsoft/node-jsonc-parser) whereas the extension-facing API is plain json 🤔 

Probably because plain json is easier to use.

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,209 @@
...
+	workspaceFolder: vscode.WorkspaceFolder,
+): Promise<void> {
+	const rbuildignoreUri = vscode.Uri.joinPath(
+		workspaceFolder.uri,
+		".Rbuildignore",
+	);
+
+	if (!(await fileExists(rbuildignoreUri))) {
+		// Do nothing if the user doesn't have one
+		return;
```
lionel-: Mention only a very small minority of R packages don't have one (i.e. non-devtools users)?

editors/code/src/command/workspace-folder-initialization.ts
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	const buffer = await vscode.workspace.fs.readFile(rbuildignoreUri);
+	const content = new TextDecoder().decode(buffer);
+
+	const newline = content.includes("\r\n") ? "\r\n" : "\n";
+	const lines = content.split(newline);
```
lionel-, line 184: ugh

editors/code/src/utils.ts
```diff
@@ -11,3 +12,20 @@ export function normalizePath(file: string): string {
...
 	return path.normalize(file);
 }
+
+/**
+ * Determine whether or not a file exists
+ *
+ * `vscode.workspace.fs.*` does not provide a way to check if a Uri exists or not,
+ * so this is supposedly the next best way to do so.
+ *
+ * This also works on directories.
```
lionel-, line 22: And URIs in general, which is the only reason to use the vscode API instead of Node?
Otherwise we'd use `fs.existsSync()`.

editors/code/src/workspace.ts
```diff
@@ -59,3 +60,134 @@ export async function getRootWorkspaceFolder(): Promise<vscode.WorkspaceFolder>
 function getWorkspaceFolders(): readonly vscode.WorkspaceFolder[] {
 	return vscode.workspace.workspaceFolders ?? [];
 }
+
+/*
+ * Select a workspace folder to act on
+ *
+ * - If 0 workspaces are open, returns `undefined` with an error message.
```
lionel-:
```suggestion
 * - If 0 workspaces are open, returns `undefined` after showing an error message.
```

editors/code/src/workspace.ts
```diff
@@ -59,3 +60,134 @@ export async function getRootWorkspaceFolder(): Promise<vscode.WorkspaceFolder>
...
+
+	// Should never get here
+	output.log(
+		`Matched a workspace folder name, but unexpectedly can't find corresponding workspace folder. Folder name: ${workspaceFolderName}.`,
+	);
+	return undefined;
+}
+
+/**
+ * Save all open dirty editor tabs relevant to the workspace folder
```
lionel-, line 131: Mention that untitled editors are not considered here.

lionel- (approved):
LG!

editors/code/src/command/workspace-folder-initialization.ts
DavisVaughan: It actually goes away on its own after about 20 seconds of waiting around

editors/code/src/command/workspace-folder-initialization.ts
DavisVaughan: Oh this comment isn't targeting non-devtool users. It's targeting people using air in a general R project!

DavisVaughan: I tweaked the comment to make it clear

editors/code/src/command/workspace-folder-initialization.ts
DavisVaughan: The fact that they use different parsers is so bonkers!

DavisVaughan: Core loop looks like

```ts
	private async doWriteConfiguration(path: JSONPath, value: any): Promise<void> {
		let content: string;
		try {
			const fileContent = await this.fileService.readFile(this.settingsResource);
			content = fileContent.value.toString();
		} catch (error) {
			if ((<FileOperationError>error).fileOperationResult === FileOperationResult.FILE_NOT_FOUND) {
				content = '{}';
			} else {
				throw error;
			}
		}

		const parseErrors: ParseError[] = [];
		parse(content, parseErrors, { allowTrailingComma: true, allowEmptyContent: true });
		if (parseErrors.length > 0) {
			throw new Error('Unable to write into the settings file. Please open the file to correct errors/warnings in the file and try again.');
		}

		const edits = this.getEdits(content, path, value);
		content = applyEdits(content, edits);

		await this.fileService.writeFile(this.settingsResource, VSBuffer.fromString(content));
	}
```

https://github.com/microsoft/vscode/blob/633f2aa00d87b257164b778ff0183ebf6036e707/src/vs/platform/configuration/common/configurationService.ts#L205

PR #401, opened 2025-08-27
Fix elided lifetimes warnings

lionel-:


DavisVaughan (approved):
What does the warning say?

lionel-:
Just that omitting unused lifetimes is confusing, so we should be explicit.

Issue #382, opened 2025-07-13
Skip tribble calls by default?


petervangalen:
Hi,
I'm using Positron and Air version 0.14.0. I like how it's fast and consistent and I'm getting used to the longer formatting. There are situations where it feels wrong, here's one:

```
id_map <- tribble(
  ~OLD_ID, ~NEW_ID,
  "P05", "P20",
  "P07", "P22",
  "P08", "P23",
  "P09", "P30",
  "P10", "P31",
  "P12", "P33"
)
```

Becomes
```
id_map <- tribble(
  ~OLD_ID,
  ~NEW_ID,
  "P05",
  "P20",
  "P07",
  "P22",
  "P08",
  "P23",
  "P09",
  "P30",
  "P10",
  "P31",
  "P12",
  "P33"
)
```

I am now avoiding formatting of this line with `# fmt: skip`
Idk if certain functions have default exceptions but if so, this would be a good one.

Issue #113, opened 2024-12-17
Tabular formatting


lionel-:
A comment like `# fmt: tabular` would instruct Air to format the next expression (which should be a function call) with these rules:

- The first line of arguments determines the number of columns
- The remaining arguments are wrapped according to the number of columns
- Arguments are aligned at commas

This would help with tribble-like data entry but could also be used with `list()` or `c()` depending on the use case.

Difficulties:

- Investigate how to hook into Biome's infra to add this special formatting

- The formatter may move comments around if they are in peculiar cases, and then a second round of formatting might cause the tabular comment to apply to something else. That's undesirable because we generally want the formatter to be idempotent.

PR #413, opened 2025-10-10
Add tabular formatting

lionel-:
Closes https://github.com/posit-dev/air/issues/113
Closes https://github.com/posit-dev/air/issues/382

### DIFF
--- a/crates/workspace/src/settings.rs
+++ b/crates/workspace/src/settings.rs
@@ -58,7 +58,7 @@ impl Default for FormatSettings {
             default_exclude: Some(Default::default()),
             default_include: Some(Default::default()),
             skip: Default::default(),
-            table: Default::default(),
+            table: Some(DEFAULT_TABLE.clone()),
         }
     }
 }
--- a/crates/workspace/src/toml_options.rs
+++ b/crates/workspace/src/toml_options.rs
@@ -14,7 +14,7 @@
 
 use std::path::Path;
 
-use crate::settings::DEFAULT_TABLE_NAMES;
+use crate::settings::DEFAULT_TABLE;
 use crate::settings::DefaultExcludePatterns;
 use crate::settings::DefaultIncludePatterns;
 use crate::settings::ExcludePatterns;
@@ -225,16 +225,12 @@ impl TomlOptions {
         let format = self.format.unwrap_or_default();
 
         let table = if format.default_table.unwrap_or(true) {
-            let mut updated: Vec<String> =
-                DEFAULT_TABLE_NAMES.iter().map(|&s| s.to_string()).collect();
-
-            if let Some(user_table) = format.table {
-                updated.extend(user_table.as_slice().iter().cloned());
-            }
-
-            Some(Table::new(updated))
+            Some(match format.table {
+                Some(table) => table.merge(&DEFAULT_TABLE),
+                None => DEFAULT_TABLE.clone(),
+            })
         } else {
-           format.table
+            format.table
         };
 
         let format = FormatSettings {
--- a/crates/settings/src/table.rs
+++ b/crates/settings/src/table.rs
@@ -20,6 +20,12 @@ impl Table {
     pub fn contains(&self, name: &str) -> bool {
         self.0.contains(name)
     }
+
+    pub fn merge(&self, table: &Table) -> Table {
+        let mut names = self.as_slice().to_owned();
+        names.extend_from_slice(table.as_slice());
+        Table::new(names)
+    }
 }
 
 impl fmt::Display for Table {
--- a/crates/workspace/src/settings/default_table.rs
+++ b/crates/workspace/src/settings/default_table.rs
@@ -1,1 +1,10 @@
-pub static DEFAULT_TABLE_NAMES: &[&str] = &["tribble", "fcase"];
+use std::sync::LazyLock;
+
+use settings::Table;
+
+static DEFAULT_TABLE_NAMES: &[&str] = &["tribble", "fcase"];
+
+pub static DEFAULT_TABLE: LazyLock<Table> = LazyLock::new(|| {
+    let names: Vec<String> = DEFAULT_TABLE_NAMES.iter().map(|&s| s.to_string()).collect();
+    Table::new(names)
+});


crates/air_r_formatter/src/r/auxiliary/argument.rs
```diff
@@ -7,32 +7,36 @@ use biome_formatter::write;
 pub(crate) struct FormatRArgument;
 impl FormatNodeRule<RArgument> for FormatRArgument {
     fn fmt_fields(&self, node: &RArgument, f: &mut RFormatter) -> FormatResult<()> {
-        let RArgumentFields { name_clause, value } = node.as_fields();
+        fmt_argument_fields(node, f)
+    }
+}
+
+pub(crate) fn fmt_argument_fields(node: &RArgument, f: &mut RFormatter) -> FormatResult<()> {
```
lionel-, line 14: This has been extracted in a method so that we can format arguments of kind Other without comments (see `parse_other()`).

xtask/codegen/src/snapshots/xtask_codegen__r_json_schema__tests__schema_can_be_generated_and_hasnt_changed.snap
```diff
@@ -33,6 +33,14 @@ expression: schema
             "null"
           ]
         },
+        "default-table": {
+          "title": "Use Air's defaults for `table`",
+          "description": "By default, `fcase()` and `tribble()` are formatted as tables. You can\n disable this default by setting this option to `false`.",
```
DavisVaughan: Can we make this match the docs for `default-exclude` more?

```
Whether or not to use defaults for `table`
```

And then in the description

```
Air automatically formats a default set of function calls as tables. If this option is set to `false`, these will be formatted as typical function calls instead. The default set includes:

- `tibble::tribble()`
- `data.table::fcase()`
```

I think it is nice to be as consistent as possible

crates/settings/src/sorted_names.rs
```diff
@@ -0,0 +1,67 @@
...
+/// object, and [Arc] provides a special `From<Vec<T>> for Arc<[T]>` for exactly this use
+/// case, which we invoke in [Skip::new()].
+///
+/// # Safety
+///
+/// This vector is sorted at creation, for use with binary search during lookups.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(description = ""))]
+pub struct SortedNames(Arc<[String]>);
```
DavisVaughan: I think we should call this `SortedStrings`? And update all usage of `names` or `name` to `strings` and `string`?

crates/air_r_formatter/src/context.rs
```diff
@@ -83,6 +85,9 @@ pub struct RFormatOptions {
 
     /// The set of functions that are skipped without requiring a `# fmt: skip` comment.
     skip: Option<Skip>,
+
+    /// The set of functions that are formatted as table without requiring a `# fmt: table` comment.
```
DavisVaughan:
```suggestion
    /// The set of functions that are formatted as tables without requiring a `# fmt: table` comment.
```

crates/air_r_formatter/src/context.rs
```diff
@@ -194,6 +212,12 @@ impl fmt::Display for RFormatOptions {
...
             }
-        )
+        )?;
+        match &self.table {
+            Some(table) => {
+                writeln!(f, "Table: {table}",)
+            }
+            None => Ok(()),
+        }
     }
```
DavisVaughan: Can you please match the `Skip: {}` formatting? i.e. it should always display `Table: None` if there are no table options used.

If we feel differently, we should change both, but the inconsistency right now isn't ideal

crates/workspace/src/settings.rs
```diff
@@ -74,7 +79,8 @@ impl FormatSettings {
             .with_line_ending(self.line_ending.finalize(source))
             .with_line_width(self.line_width)
             .with_persistent_line_breaks(self.persistent_line_breaks)
-            // Note that `clone()` here is on an `Arc`
+            // Note that `clone()` on these options is on an `Arc`
             .with_skip(self.skip.clone())
+            .with_table(self.table.clone())
```
DavisVaughan:
```suggestion
            .with_skip(Arc::clone(self.skip))
            .with_table(Arc::clone(self.table))
```

Oh hey I think this is a much better option! I believe the recommendation is to use `Arc::clone()` at call sites to convey that it is cheap. Then we don't need the comment.

crates/workspace/src/toml_options.rs
```diff
@@ -194,12 +196,43 @@ pub struct FormatTomlOptions {
     /// of `tibble::tribble()` calls. In the long term, we may have more sophisticated
```
DavisVaughan: Please remove these docs now! If you search for `In the short term` then you will probably find another place in the codebase where this can be removed

crates/workspace/src/toml_options.rs
```diff
@@ -194,12 +196,43 @@ pub struct FormatTomlOptions {
...
+    /// # Function calls to format as tables
+    ///
+    /// Some variadic functions like `tibble::tribble()` or
+    /// `data.table::fcase()` are meant to be formatted as tables. Air
+    /// helps you do it automatically with the `# fmt: table` directive.
+    /// Use this setting to determine which functions should automatically be
+    /// formatted as tables without requiring a comment directive.
+    ///
+    /// By default, `fcase()` and `tribble()` are formatted as tables. You can
+    /// disable this default by setting the `default-table` option to `false`.
```
DavisVaughan:
```suggestion
    /// Some function calls are meant to be formatted as tables, rather than being
    /// formatted in a flat or expanded layout. For a single one-off function call,
    /// you can use a `# fmt: table` comment to request a table layout. For function
    /// calls that you use a lot, use this setting to add them to a list of function
    /// calls that are automatically formatted as tables without requiring a
    /// `# fmt: table` comment.
    ///
    /// For example, using `table = ["my_table"]` would automatically format calls
    /// to `my_table()` in a table layout.
    ///
    /// See `default-table` for the list of function calls that are automatically
    /// formatted as tables by default.
```

Again trying to mimic `skip` a bit more. I feel like it is worth not mentioning `tribble()` and `fcase()` here? We list them in `default-table`. I feel like mentioning them here as examples of function you might want to table format would cause people to actually list them here, not realizing that we already format them this way by default.

crates/workspace/src/toml_options.rs
```diff
@@ -194,12 +196,43 @@ pub struct FormatTomlOptions {
...
+        if format.default_table.unwrap_or(true) {
+            let mut updated: Vec<String> =
+                DEFAULT_TABLE_NAMES.iter().map(|&s| s.to_string()).collect();
+
+            if let Some(user_table) = table {
+                updated.extend(user_table.as_slice().iter().cloned());
+            }
+
+            table = Some(Table::new(updated));
+        }
```
DavisVaughan: I actually do like that all of the setting transformations are inlined right now, I think this would work well inlined too

```rust
            table: if format.default_table.unwrap_or(true) {
                let mut everything: Vec<String> =
                    DEFAULT_TABLE_NAMES.iter().map(|&s| s.to_string()).collect();

                if let Some(table) = format.table {
                    everything.extend(table.as_slice().iter().cloned());
                }

                Some(Table::new(everything))
            } else {
                format.table
            },
```

crates/air_r_formatter/tests/specs/r/air.toml
```diff
@@ -0,0 +1,2 @@
+[format]
+exclude=["**/*.R"]
```
DavisVaughan: I think you should remove this file

Currently Air does not recurse into folders to find `air.toml`, so having this here has no effect unless you literally open the VS Code project at `tests/specs/r/`, which we don't do

So I think it is just confusing because saving in the R files here still rearranges them

crates/air_r_formatter/tests/specs/r/table/table-default.R
```diff
@@ -0,0 +1,33 @@
+#' [format]
```
DavisVaughan, line 1: Remove?

crates/air_r_formatter/tests/specs/r/skip/skip-tribble2.R
DavisVaughan, line 1: I think this whole file and `skip-tribble2.R.snap` are mistakes? We don't skip tribbles, we table them. And we have separate table tests.

crates/air_r_formatter/tests/specs/r/call_table.R
```diff
@@ -0,0 +1,366 @@
...
+foooooooooo,baaaaaaaaar,foooooooooo,baaaaaaaaar,foooooooooo,baaaaaaaaar,foooooooooo,baaaaaaaaar,foooooooooo(baaaaaaaaar,foooooooooo,baaaaaaaaar),
+1+100, 1+1000,
+)
+
+# ------------------------------------------------------------------------
+# Comments
+
+list(# comment1
+# comment2
+)# comment3
```
DavisVaughan, line 100: Remove this or add `# fmt: table`?

crates/air_r_formatter/tests/specs/r/call_table.R
```diff
@@ -0,0 +1,366 @@
...
+    1+1, 1+1
+  )
+
+# Fallback syntax for these cases
+foo %=%
+  # fmt: table
+  list(
+    1+1, 1+1,
+    1+1, 1+1
+  )
```
DavisVaughan, line 238: I think this feels like a reasonable restriction!

crates/air_r_formatter/tests/specs/r/call_table.R.snap
```diff
@@ -0,0 +1,783 @@
...
+    3.456 ,
+)
+
+# Mixed integers and decimals
+# fmt: table
+list(
+  1000     ,
+     2.5   ,
+   50L     ,
+   123.456 ,
```
DavisVaughan: I believe we said we wanted `L` to align on the decimal?

I'd also expect these to align like this

```
50L
 1
```

i.e. the `L` is a pseudo-decimal when you mix integers and non-fractional doubles

crates/air_r_formatter/tests/specs/r/call_table.R.snap
```diff
@@ -0,0 +1,783 @@
...
+list(
+      1.10 ,
+  "12"     ,
+      1    ,
+     2L    ,
+  "3"      ,
+  10000    ,
+      3.5  ,
+  f()
+)
```
DavisVaughan, line 767: I do wonder if in the future (NOT now) we will want to revise this to all left-aligned everything to account for stuff like this. I do like that the mixed types get called out right now though, I just wonder if there are legit usages of something like this, like `fcase()` with `stop()`.

But maybe the benefit of right aligned numerics far outweighs any weirdness in some less common cases.

crates/air_r_formatter/tests/specs/r/call_table.R.snap
```diff
@@ -0,0 +1,783 @@
...
+
+# ------------------------------------------------------------------------
+# Unary operators
+
+# fmt: table
+list(
+    -1.200 ,
+   +2L     ,
```
DavisVaughan: I expect this to adjust when we fix `L` alignment

crates/air_r_formatter/tests/specs/r/call_table.R
```diff
@@ -0,0 +1,366 @@
...
+
+# fmt: table
+list(
+{ foo },  2
+)
+
+# ------------------------------------------------------------------------
+# Named argument
+
+# Typical case
```
DavisVaughan, line 386:
```suggestion
# Named argument

# The first named argument stops the table. That argument and all
# that follow it are laid out in expanded form.

# Typical case
```

crates/air_r_formatter/tests/specs/r/call_table.R
```diff
@@ -0,0 +1,366 @@
...
+)
+
+# fmt: table
+list(
+  1,2,
+  3,4,
+
+  # comment
+  foo = 1, 2, 3
+)
```
DavisVaughan, line 409: Would you mind adding a test for "removes line breaks"? Something like

```
# fmt: table
list(
  c(
    1, 2
  ), 3,
  c(4, 5), 6,
  c(
    7, 8), 9
)
```

It's nice to document somewhere that opting in to table format means you lose all line breaks.

It's nice that row detection seems to work right here still

crates/air_r_formatter/src/r/auxiliary/call.rs
```diff
@@ -1,52 +1,104 @@
...
+use biome_rowan::AstNode;
 use biome_rowan::SyntaxResult;
+use comments::Directive;
+use comments::FormatDirective;
 use settings::Skip;
 
 #[derive(Debug, Clone, Default)]
-pub(crate) struct FormatRCall;
+pub(crate) struct FormatRCall {
+    table: bool,
```
DavisVaughan: Would you mind making this `table: Option<bool>`? Imagine if a caller tried to set this option by explicitly supplying `table: false`. That would not work as expected, because `is_table_call()` would still run rather than just accepting what the caller explicitly provided.

I also find it hard to grok that `table = false` is the `Default` that we get when this is unset.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
```diff
@@ -19,16 +19,51 @@ use air_r_syntax::RSubset2Arguments;
...
+
+                // Table formatting failed, fall back to verbatim. Ideally
+                // we'd emit diagnostics about why tabular formatting failed
+                // here.
+                write!(f, [format_verbatim_node(node.syntax())])
+            }
+        } else {
+            self.fmt_call_like(node, f)
+        }
     }
```
DavisVaughan, line 67:
```suggestion
    fn fmt_fields(&self, node: &RCallArguments, f: &mut RFormatter) -> FormatResult<()> {
        if self.table && !node.items().is_empty() {
            // Special table formatting
            let snapshot = f.state_snapshot();

            if let Some(()) = self.fmt_table(node, f)? {
                return Ok(());
            } else {
                f.restore_state_snapshot(snapshot);

                // Table formatting failed, fall back to verbatim. Ideally
                // we'd emit diagnostics about why tabular formatting failed
                // here.
                write!(f, [format_verbatim_node(node.syntax())])?;
                return Ok(());
            }
        }

        RCallLikeArguments::Call(node.clone()).fmt(f)
    }
```

I tend to prefer this for a few reasons:

- `!node.items().is_empty()` moves into this function, so `fmt_table()` is now never expected to bail for this reason, and does not know about `fmt_call_like()`
- No need for `fmt_call_like()` at all, we get to inline it again
- That's nice because now the `node.clone()` in `RCallLikeArguments::Call(node.clone()).fmt(f)` is no longer hidden
- The "special" case of table formatting looks like a special early exit. This will be nice if we end up with additional special cases in the future.

If you do this, also remove `fmt_call_like()` and then remove it from `call_arguments_table.rs` as well, which cleans things up pretty nicely!

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -184,24 +186,29 @@ fn fmt_binary_sticky(
...
     operator: SyntaxToken<RLanguage>,
     right: AnyRExpression,
     f: &mut Formatter<RFormatContext>,
 ) -> FormatResult<()> {
-    let right = format_with(|f| {
+    // Check for table directive here to simplify lifetimes with
+    // `format_assignment_rhs()`
+    let table = is_table_by_comment(node, f);
+
+    let right_formatted = format_with(|f| {
```
DavisVaughan: I'd argue this should still be called `right` because it hasn't really been formatted yet, its like a delayed closure

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -211,11 +218,23 @@ fn fmt_binary_assignment(
...
+    format_with(move |f| {
+        if table {
+            if let AnyRExpression::RCall(call) = expr {
+                let options = FormatRCallArgumentsOptions { table: true };
+                return write!(f, [call.format().with_options(options)]);
+            }
+        }
+        write!(f, [expr.format()])
+    })
+}
```
DavisVaughan:
```suggestion
fn format_assignment_rhs(right: &AnyRExpression, table: bool) -> impl Format<RFormatContext> {
    format_with(move |f| {
        if table {
            if let AnyRExpression::RCall(call) = right {
                let options = FormatRCallArgumentsOptions { table: true };
                return write!(f, [call.format().with_options(options)]);
            }
        }
        write!(f, [right.format()])
    })
}
```

Because `binary_assignment_has_persistent_line_break()` also uses `right`, so it's slightly more consistent I think, and a little easier to read on its own

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+    let mut cols: Vec<ColumnInfo> = Vec::new();
+    let mut rows: Vec<Vec<ArgData>> = Vec::new();
+    let mut current_row = Vec::new();
+
+    let mut items = args.elements().enumerate();
+    let mut remaining = vec![];
+
+    for (i, arg) in &mut items {
+        // We've encountered an `off` comment before, keep collecting remaining args
+        if remaining.len() > 0 {
```
DavisVaughan: There are quite a few clippy warnings in here. I can handle them if you can't see them, but we should resolve them all before merging

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
+use crate::r::auxiliary::call_arguments::FormatRCallArguments;
```
DavisVaughan, line 1: I think we talked about the fact that a user's explicit empty lines were going to be retained, but that doesn't seem to be the case

```r
# fmt: table
list(
  1 , 2 ,
  
  3 , 4
)
```

This removes the blank line, not sure if that is by design or not (needs a test either way?)

If we do want to retain explicit blank lines, I'd argue we should do the same as normal function calls where:

```r
# This gets removed
# fmt: table
list(

  1 , 2 ,
  3 , 4
)

# This gets removed
# fmt: table
list(
  1 , 2 ,
  3 , 4

)

# This is allowed
# fmt: table
list(
  1 , 2 ,

  3 , 4
)

# This is trimmed back to 1 empty line
# fmt: table
list(
  1 , 2 ,

  
  3 , 4
)
```

crates/air_r_formatter/src/r/auxiliary/call.rs
```diff
@@ -1,52 +1,104 @@
...
+        self.table = options.table;
+        self
     }
 }
 
 #[inline]
-fn is_suppressed_by_skip(node: &RCall, f: &RFormatter) -> SyntaxResult<bool> {
-    let Some(skip) = f.options().skip() else {
-        // Nothing to do if user didn't supply any skip functions
+fn is_skip_by_option(node: &RCall, f: &RFormatter) -> SyntaxResult<bool> {
```
DavisVaughan: I have some alternative names I'd like to pair with you on regarding these functions, I find them a bit confusing right now but I think I have a nice way to clarify them all

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+                    write!(f, [hard_line_break()])?;
+                }
+
+                for (col_j, arg_data) in row.iter().enumerate() {
+                    let (left_pad, right_pad) = if column_info[col_j].max_width > 0 {
+                        column_info[col_j].padding(&arg_data.kind)
+                    } else {
+                        // Empty columns are not padded, so that commas stick to
+                        // each other as in regular calls: `list(,,,)`
+                        (0, 0)
```
DavisVaughan: Is this necessary anymore?

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+
+                            // 0-length arguments are holes. Don't print them
+                            // because a `text("")` after a `Space` will prevent
+                            // the latter from being considered trailing by the
+                            // printer, and won't be removed if trailing.
+                            if !text.is_empty() {
+                                write!(f, [dynamic_text(text, 0.into())])?;
+                            }
+
+                            format_trailing_comments(arg_data.node.syntax()).fmt(f)?;
```
DavisVaughan:
```suggestion
                            let arg_syntax = arg_data.node.syntax();

                            // Suppression comments do nothing inside a table
                            f.comments().mark_suppression_checked(arg_syntax);

                            // We've formatted the argument without comments, so
                            // we're in charge of formatting them
                            format_leading_comments(arg_syntax).fmt(f)?;

                            // 0-length arguments are holes. Don't print them
                            // because a `text("")` after a `Space` will prevent
                            // the latter from being considered trailing by the
                            // printer, and won't be removed if trailing.
                            if !text.is_empty() {
                                write!(f, [dynamic_text(text, TextSize::from(0))])?;
                            }

                            format_trailing_comments(arg_syntax).fmt(f)?;
```

- Inline `disable_skip_comments()` with appropriate comment
- Consistently use `arg_syntax`

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+    if count > 0 {
+        write!(f, [dynamic_text(&" ".repeat(count), 0.into())])?;
+    }
+    Ok(())
+}
+
+/// Skip comments nested in the table are invalid, but mark them as checked so the formatter
+fn disable_skip_comments(syntax: &RSyntaxNode, f: &RFormatter) {
+    f.context().comments().mark_suppression_checked(syntax);
+}
```
DavisVaughan: Ok I now understand what this is trying to do - we just need to say "hey, we see that this node could have comments, we assert that we ignore any suppression comments while inside the table"

Elsewhere when we do this we just inline the call to `f.comment().mark_suppression_checked(node)` directly (note you can drop `.context()`), and that is what I'd suggest we do here as well to be less confusing.

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+fn build_table_impl(args: &RArgumentList, f: &mut RFormatter) -> FormatResult<Option<TableInfo>> {
+    let mut cols: Vec<ColumnInfo> = Vec::new();
+    let mut rows: Vec<Vec<ArgData>> = Vec::new();
+    let mut current_row = Vec::new();
+
+    let mut items = args.elements().enumerate();
+    let mut remaining = vec![];
+
+    for (i, arg) in &mut items {
+        // We've encountered an `off` comment before, keep collecting remaining args
```
DavisVaughan:
```suggestion
        // We've encountered a named argument before, keep collecting remaining args
```

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+        // should be unnecessary as we grow row elements one by one, it is just
+        // to be safe.
+        while cols.len() <= column_index {
+            cols.push(ColumnInfo::default());
+        }
+
+        let kind = match ArgKind::parse(arg_node, f)? {
+            Some(kind) => kind,
+            None => return Ok(None),
+        };
```
DavisVaughan:
```suggestion
        let Some(kind) = ArgKind::parse(arg_node, f)? else {
            // Argument has a forced newline, bail on table formatting
            return Ok(None);
        };
```

Had to go look up what it means to hit the `None` case, so it felt like it could use a comment

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+                Some(Self::parse_decimal(value)?)
+            }
+            Some(AnyRExpression::RUnaryExpression(value)) => Self::parse_unary(arg, value, f)?,
+            Some(AnyRExpression::AnyRValue(AnyRValue::RBogusValue(_))) => {
+                return Err(FormatError::SyntaxError);
+            }
+            _ => Self::parse_other(arg, f)?,
+        };
+
+        Ok(kind)
```
DavisVaughan:
```suggestion
        match arg.value() {
            Some(AnyRExpression::AnyRValue(AnyRValue::RIntegerValue(value))) => {
                Ok(Some(Self::parse_integer(value)?))
            }
            Some(AnyRExpression::AnyRValue(AnyRValue::RDoubleValue(value))) => {
                Ok(Some(Self::parse_decimal(value)?))
            }
            Some(AnyRExpression::RUnaryExpression(value)) => Ok(Self::parse_unary(arg, value, f)?),
            Some(AnyRExpression::AnyRValue(AnyRValue::RBogusValue(_))) => {
                Err(FormatError::SyntaxError)
            }
            _ => Ok(Self::parse_other(arg, f)?),
        }
```

?

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -0,0 +1,508 @@
...
+        };
+
+        Ok(ArgKind::Numeric {
+            integer_width,
+            fractional_width, // Note this might be 0, e.g. `1.`
+        })
+    }
+
+    fn parse_other(arg: &RArgument, f: &mut RFormatter) -> FormatResult<Option<ArgKind>> {
+        disable_skip_comments(arg.syntax(), f);
```
DavisVaughan: I believe you can remove this! You mark suppression comments as checked in the formatting step (which feels like the correct place anyways), so right now they are being marked twice.

That gives even more value to just inlining this function.

docs/configuration.qmd
```diff
@@ -263,3 +265,12 @@ igraph::graph_from_literal(A +-+ B +---+ C ++ D + E)
 
 In the short term, we also anticipate that this will be useful to avoid formatting of `tibble::tribble()` calls.
 In the long term, Air may have more sophisticated features for automatically formatting using a specified alignment.
```
DavisVaughan: Remove this!

docs/configuration.qmd
```diff
@@ -263,3 +265,12 @@ igraph::graph_from_literal(A +-+ B +---+ C ++ D + E)
...
+
+### table and default-table {#configuration-table}
+
+The set of function names to format as tables, even without a `# fmt: table` comment.
+
+A list of strings, i.e. `table = ["my_table_function"]`.
+
+By default, Air always formats calls to `tribble()` (from tibble) and `fcase()` (from data.table) as tables.
+To disable these defaults, set `default-table` to `false`.
```
DavisVaughan: I have tried fairly hard to keep this and `toml_options.rs` in sync

I think we should:

- Split `table` and `default-table` into two separate sections like `exclude` and `default-exclude`
- Copy and paste over the docs from `toml_options.rs` after making my other edits I suggested there and try to keep them as synced as possible. Ideally a rust macro would be involved here, I just haven't been able to make one yet.

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
+With `# fmt: skip` we've already seen one kind of comment directive, but that's not all.
+Air supports another directive, `# fmt: table`, and we plan to add more in the future.
+
+In general, comment directives are a way to opt out of Air's default automatic formatting and opt into specialized format layouts.
+They generally apply to the expression below the comment, with some possible nuances.
+
+- `skip`: Applies to any following expression.
+- `table`: Applies to the next expression if a function call (e.g. `foo()` or an assigned function call (e.g. `foo <- bar()`)).
+
+# Table formatting
```
DavisVaughan:
```suggestion
## Table formatting
```

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
-In the long term, Air may provide more sophisticated tooling for formatting using a specified alignment.
+# Comment directives
+
+With `# fmt: skip` we've already seen one kind of comment directive, but that's not all.
+Air supports another directive, `# fmt: table`, and we plan to add more in the future.
+
+In general, comment directives are a way to opt out of Air's default automatic formatting and opt into specialized format layouts.
+They generally apply to the expression below the comment, with some possible nuances.
+
+- `skip`: Applies to any following expression.
```
DavisVaughan:
```suggestion
- `skip`: Applies to the next expression.
```

Symmetric with what you say next

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
+- `table`: Applies to the next expression if a function call (e.g. `foo()` or an assigned function call (e.g. `foo <- bar()`)).
+
+# Table formatting
+
+Note: This feature is experimental. We'd be grateful for any feedback!
```
DavisVaughan:
```suggestion
::: callout-caution
## This feature is experimental. We'd be grateful for any feedback!
:::
```

The Quarto Callout feature looks quite nice here

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
 skip = ["graph_from_literal"]
 ```
```
DavisVaughan, line 618: Can you please open this file in the Positron Quarto Visual Editor and save it there? It causes a number of diffs from the current state.

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
+Note: This feature is experimental. We'd be grateful for any feedback!
+
+This directive is particularly useful with calls like tibble's `tribble()` or data.table's `fcase()`.
+It formats calls taking unnamed arguments via `...` in a tabular way.
+
+- Commas are aligned, creating column guides
+- Numeric values are right-aligned, making them easier to read and compare
+- Decimals are aligned, making it easier to distinguish integer and fractional parts
+- Everything else is left-aligned, making it easier to read from left to right
+- Complex calls are formatted in a flat layout, avoiding line breaks where possible.
```
DavisVaughan:
```suggestion
- Commas are aligned, creating column guides
- Numeric values are right-aligned
- Everything else is left-aligned
- Decimals are aligned to distinguish integer and fractional parts
- Complex calls are formatted in a flat layout, avoiding line breaks
```

It feels like we can trim this down a good bit and still get the point across

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
+
+- The layout of the table is never changed by the formatter.
+  It never introduces line breaks that don't already exist to avoid messing up the tabular data, which could be hard to recover from.
+- Tables are allowed to be jagged with different numbers of columns per rows.
+  This allows formatting partial tables as you write them.
+
+Note that if a complex expression contains a line break that Air can't flatten out, tabular formatting is disabled.
+You'll have to find the offending argument and make it flat to turn table formatting back on.
+
+By default, all calls to `tribble()` and `fcase()` are formatted as tables, even without a `fmt: table` directive.
```
DavisVaughan:
```suggestion
By default, all calls to `tribble()` and `fcase()` are formatted as tables, even without a `# fmt: table` directive.
```

docs/formatter.qmd
```diff
@@ -616,5 +616,53 @@ Specifying this would retain the original formatting of the `graph_from_literal(
...
+Two [configuration options](configuration.qmd#configuration-table) help you tweak this default behavior:
+
+ - `table`: Extend the default table calls (e.g. `tribble()`) with custom function names that should be formatted as tables across your whole project without requiring a `fmt: table` directive.
```
DavisVaughan:
```suggestion
 - `table`: Extend the default table calls (e.g. `tribble()`) with custom function names that should be formatted as tables across your whole project without requiring a `# fmt: table` directive.
```

crates/air_r_formatter/src/context.rs
lionel-: I did this on purpose as I thought settings like Table (we might have more in the future) are less important and don't need to be distracting here. Let's discuss when we pair.

crates/workspace/src/settings.rs
lionel-: oh nice

lionel-: That doesn't work because the Arc is deeply nested, here we are cloning Options technically.

lionel-: I updated the comment:

```rs
            // Note that `clone()` on these options is ultimately on an `Arc`
```

crates/workspace/src/toml_options.rs
lionel-: I think it actually works better as I wrote it because it makes it clearer what's the common part.

Since I don't really mind either way, I updated with your suggestion.

crates/air_r_formatter/tests/specs/r/air.toml
lionel-: hmm I must have gotten confused and disabled formatting in my setup for these files in two different ways

crates/air_r_formatter/tests/specs/r/table/table-default.R
lionel-: This is needed for the default configuration to apply in the snapshots

crates/air_r_formatter/tests/specs/r/call_table.R.snap
lionel-: My current position is that the benefits far outweigh the weirdness of mixed types, which should be very rare in practice compared to e.g. a mix of calls and numerics

crates/air_r_formatter/src/r/auxiliary/call.rs
lionel-: oh very good point!

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
lionel-: I'll make it `right_format`. While it hasn't been formatted yet, it's been converted to an object that implements `Format`.

crates/air_r_formatter/src/r/auxiliary/call_arguments.rs
lionel-: I don't mind the early return style but I think I prefer handling the special case in the table implementation. That we delegate to CallLikeArguments is an implementation detail.

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
lionel-: yes but the comment is now slightly misleading, it's now:

```rs
                        // For empty columns don't add any incompressible whitespace
```

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
lionel-: yep nice

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
lionel-: yep, can even remove some `?` unwrapping then

docs/formatter.qmd
lionel-: hmm it's not really meant to be symmetric, `skip` applies to literally any following expression, but  that's not the case for `table`

docs/formatter.qmd
lionel-: hmm

docs/configuration.qmd
lionel-: gotcha

crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
lionel-: I can't get them to show up in zed but I see them with `cargo clippy`

lionel-:
Alright! I think this is now ready for a last round of pairing

crates/air_r_formatter/src/context.rs
DavisVaughan: We decided to make `Skip` optional here too

Issue #350, opened 2025-05-29
Consider limiting format directive comments to own-line comments


DavisVaughan:
TLDR is that I think we should respect

```r
# fmt: skip
fn(
 a, b
)
```

but ignore (and possibly lint)

```r
fn(
 a, b # fmt: skip
)
```

I have many feelings about this, but the overarching rationale is that it is _very_ hard to have precise semantic meaning with end-of-line comments, and it is much easier to have precise semantic meaning with own-line comments that _lead a node_.

This would be much more similar to how rustfmt works, because their skip attribute is _node based_. They simply cannot have end-of-line attributes, so they don't have to think about this at all.

```rust
#[rustfmt::skip]
fn()
```

---

There is a very very good summary of this on the ruff side, where they have basically come to the same conclusion
https://github.com/astral-sh/ruff/discussions/6670

In particular, https://github.com/astral-sh/ruff/discussions/6670#discussioncomment-6788116

> are there any lessons we should takeaway for any suppression comment redesigns we explore in the future?

> Don't use trailing end-of-line comments. I don't have too strong evidence because I haven't looked as much into the other comments, but they seem the hardest to preserve. My impression (and prettier's recommendation) is to use leading own line comments, ideally, such that are node-based (they suppress the next node and not the next line).

--- 

Here is Prettier's advice:

https://prettier.io/docs/next/rationale#comments

> Generally, you get the best results when placing comments on their own lines, instead of at the end of lines. Prefer // eslint-disable-next-line over // eslint-disable-line.

---

I think we should strongly consider the possibility of limiting our "magic" comments to own-line only, and try and make them as "node based" as possible, as this is likely going to make our lives a lot easier, and behavior more predicable for users.

DavisVaughan:
This is also related to https://github.com/posit-dev/air/issues/256, and should be kept in mind if we use Air infra to back some other linter in the future.

PR #419, opened 2025-10-16
Restrict comment directives to leading comments

lionel-:
And only in expression and argument lists for now (similar to how we handle section comments in Ark).

Closes https://github.com/posit-dev/air/issues/350

### DIFF
--- a/crates/air_r_formatter/src/lib.rs
+++ b/crates/air_r_formatter/src/lib.rs
@@ -261,7 +261,7 @@ where
 ///
 /// We intentionally only consider directives in leading comments. This is a
 /// departure from Biome (and Ruff?).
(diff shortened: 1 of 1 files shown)


crates/air_r_formatter/src/lib.rs
```diff
@@ -189,6 +193,9 @@ where
 pub(crate) type RFormatter<'buf> = Formatter<'buf, RFormatContext>;
 
 /// Rule for formatting an R [AstNode].
+///
+/// This is one of the traits that are expected by the Biome-generated code.
```
DavisVaughan:
```suggestion
/// This is one of the traits that is expected by the Biome-generated code.
```

PR #423, opened 2025-10-17
Use a `CommentDirectives` trait to make `directives()` feel ergonomic

DavisVaughan:


crates/air_r_formatter/src/directives.rs
```diff
@@ -1,55 +1,64 @@
...
+use biome_rowan::SyntaxResult;
 use comments::{Directive, FormatDirective};
 use settings::Skip;
+use settings::Table;
 
-/// Returns `true` if the node has a suppression comment and should use the same formatting as in the source document.
+use crate::RFormatter;
+
+/// Generic trait for extracting [comments::Directive]s from a node's comments
+pub trait CommentDirectives {
```
lionel-, line 14: I think I'd prefer a `CommentsExt` trait that makes it clearer that this is an extension trait instead of a generic trait.

PR #424, opened 2025-10-17
Move `mark_suppression_checked()` out of `has_skip_directive()`

DavisVaughan:
I set out to remove `mark_suppression_checked()` from `has_skip_directive()` because, as discussed, it was weird for this particular predicate function to be marking as well.

I realized that we actually had a footgun in the process.

In https://github.com/posit-dev/air/pull/419 we switched away from using `f.context().comments().is_suppressed()` to instead use our directive based approach.

That `Comments::is_suppressed()` method we switched away from does a few things that are noteworthy. It:

- Calls `mark_suppression_checked()`
- Runs on ALL OF leading, dangling, and trailing comments
- Calls `CommentStyle::is_suppression()` with `text` of the comment to determine if it is a suppression comment or not

Our directive based approach:

- Calls `mark_suppression_checked()`
- Runs on JUST leading comments

Even after #419, we still had an implementation for `is_suppression()` floating around that parsed the text and returned `true` if it was a `Skip` directive - regardless of whether it was a leading, trailing, or dangling comment, even though we only want to consider leading ones!

That means that if anyone called `f.context().comments().is_suppressed()`, then it would incorrectly report that a node was suppressed even if it has a non-leading skip directive. This is the footgun, and in fact we did this in `unary_expression.rs`.

I've removed this footgun by making `is_suppression()` unreachable, and by adding a big comment about what you should be looking to use instead. You instead want `f.comments().has_skip_directive(node)` to determine if a node should be suppressed, combined with an explicit call to `f.comments().mark_suppression_checked(node)` on that node to declare that you've checked if it needs to be suppressed or not. That feels a lot better than having the marking tied up in `has_skip_directive()`.

---

Along the way I rejiggered the `CommentDirectives` trait to just expose two extensions

```rust
fn has_skip_directive(&self, node: &SyntaxNode<Self::Language>) -> bool;
fn has_table_directive(&self, node: &SyntaxNode<Self::Language>) -> bool;
```

Previously you'd do `f.comments().directives().any(...)` and now you just do `f.comments().has_skip_directive(node)`, which goes nicely with the rest of the `f.comments().has_*()` family like `f.comments().has_leading_comments(node)`.

---

Lastly, it's worth noting that I really do wish we could just use `f.context().comments().is_suppressed(node)`. In theory between the `node` and the `comments()` we'd have everything we need to correctly determine if it is a correct suppression comment or not. In practice this doesn't work for two reasons:

- `Comments::is_suppressed(node)` isn't overridable, and looks at all of leading, dangling, and trailing comments. This would be okay, but...
- ...`CommentStyle::is_suppression(text)`, which is overridable, only gets the `text`, so we have no way of knowing whether its leading or not.

I think ideally `is_suppressed()` itself would have been configurable in some way?

### DIFF
--- a/crates/air_r_formatter/src/comments.rs
+++ b/crates/air_r_formatter/src/comments.rs
@@ -20,27 +20,11 @@ use biome_formatter::comments::DecoratedComment;
 use biome_formatter::comments::SourceComment;
 use biome_formatter::write;
 use biome_rowan::AstNode;
-use biome_rowan::SyntaxNode;
 use biome_rowan::SyntaxTriviaPieceComments;
-use comments::Directive;
-use comments::FormatDirective;
 
-use crate::directives::CommentDirectives;
 use crate::prelude::*;
 
 pub type RComments = Comments<RLanguage>;
-
-impl CommentDirectives for RComments {
-    type Language = RLanguage;
-
-    fn directives(&self, node: &SyntaxNode<Self::Language>) -> impl Iterator<Item = Directive> {
-        // We intentionally only consider directives in leading comments. This is a
-        // departure from Biome (and Ruff?).
-        self.leading_comments(node)
-            .iter()
-            .filter_map(|c| comments::parse_comment_directive(c.piece().text()))
-    }
-}
 
 #[derive(Default)]
 pub struct FormatRLeadingComment;
@@ -65,9 +49,21 @@ pub struct RCommentStyle;
 impl CommentStyle for RCommentStyle {
     type Language = RLanguage;
 
-    fn is_suppression(text: &str) -> bool {
-        comments::parse_comment_directive(text)
-            .is_some_and(|directive| matches!(directive, Directive::Format(FormatDirective::Skip)))
+    /// Must never be called!
+    ///
+    /// Would be called by [biome_formatter::Comments::is_suppressed()], i.e. by a call
+    /// to `f.comments().is_suppressed()`. We can't use that, as it uses
+    /// [biome_formatter::Comments::leading_dangling_trailing_comments()] and will check
+    /// all three styles of comments for suppression. In Air, we only allow suppression
+    /// on leading comments.
+    ///
+    /// If you are writing your own `is_suppressed()` method for a [FormatNodeRule], you
+    /// instead want [crate::CommentDirectives::has_skip_directive()] to check if a node
+    /// should be suppressed, combined with an explicit call to
+    /// [biome_formatter::Comments::mark_suppression_checked()] to acknowledge that you've
+    /// checked that node.
+    fn is_suppression(_text: &str) -> bool {
+        unreachable!("You want `CommentDirectives::has_skip_directive()` instead");
     }
 
     fn get_comment_kind(_comment: &SyntaxTriviaPieceComments<RLanguage>) -> CommentKind {
--- a/crates/air_r_formatter/src/lib.rs
+++ b/crates/air_r_formatter/src/lib.rs
@@ -16,7 +16,7 @@ use crate::comments::RCommentStyle;
 use crate::context::RFormatContext;
 use crate::context::RFormatOptions;
 use crate::cst::FormatRSyntaxNode;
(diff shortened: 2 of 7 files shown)


crates/air_r_formatter/src/comments.rs
```diff
@@ -65,9 +49,21 @@ pub struct RCommentStyle;
...
+    /// on leading comments.
+    ///
+    /// If you are writing your own `is_suppressed()` method for a [FormatNodeRule], you
+    /// instead want [crate::CommentDirectives::has_skip_directive()] to check if a node
+    /// should be suppressed, combined with an explicit call to
+    /// [biome_formatter::Comments::mark_suppression_checked()] to acknowledge that you've
+    /// checked that node.
+    fn is_suppression(_text: &str) -> bool {
+        unreachable!("You want `CommentDirectives::has_skip_directive()` instead");
     }
```
DavisVaughan, line 67: Scary!

crates/air_r_formatter/src/lib.rs
```diff
@@ -213,7 +213,10 @@ where
 
     /// Returns `true` if the node is suppressed and should use the same formatting as in the source document.
     fn is_suppressed(&self, node: &N, f: &RFormatter) -> bool {
-        has_skip_comment(node.syntax(), f)
+        let node = node.syntax();
+        let comments = f.comments();
+        comments.mark_suppression_checked(node);
+        comments.has_skip_directive(node)
     }
```
DavisVaughan, line 220: There is now this general pattern that your `is_suppressed()` overrides have to do two things:
- Call `mark_suppression_checked()`
- Call `has_skip_directive()`

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -503,14 +502,19 @@ fn fmt_binary_chain(
...
-        if has_skip_comment(node.syntax(), f) {
-            tracing::warn!("Can't use a suppression comment partway through a binary chain.");
+        // It's only possible to suppress the formatting of the whole binary expression
+        // formatting OR the formatting of the right hand side value but not of a nested
+        // binary expression, so we aren't going to respect any skip directives here, but
+        // we must mark that we've checked them.
+        comments.mark_suppression_checked(node.syntax());
+        if comments.has_skip_directive(node.syntax()) {
+            tracing::warn!("Can't use a skip comment partway through a binary chain.");
         }
```
DavisVaughan, line 518: This does make sense to me now, and is correct

crates/air_r_formatter/src/r/auxiliary/call.rs
```diff
@@ -32,12 +31,15 @@ impl FormatNodeRule<RCall> for FormatRCall {
     }
 
     fn is_suppressed(&self, node: &RCall, f: &RFormatter) -> bool {
-        has_skip_comment(node.syntax(), f) || in_skip_setting(node, f).unwrap_or(false)
+        let comments = f.comments();
+        comments.mark_suppression_checked(node.syntax());
+        comments.has_skip_directive(node.syntax()) || in_skip_setting(node, f).unwrap_or(false)
     }
```
DavisVaughan, line 37: Again here's that general pattern of an `is_suppressed()` override doing both `mark_suppression_checked()` and `has_skip_directive()`, with an additional check for the `skip` configuration setting since this is a call

crates/air_r_formatter/src/r/auxiliary/unary_expression.rs
```diff
@@ -18,7 +19,7 @@ impl FormatNodeRule<RUnaryExpression> for FormatRUnaryExpression {
 
         let argument = format_with(|f| {
             if f.comments().has_comments(argument.syntax())
-                && !f.comments().is_suppressed(argument.syntax())
+                && !f.comments().has_skip_directive(argument.syntax())
```
DavisVaughan, line 22: This was our previous footgun, which now nicely forced a panic due to hitting the `unreachable!()` path I added

Issue #405, opened 2025-09-16
Support for VS Code Jupyter Notebooks


kv9898:
Currently, air formats olain R documents like a charm! But it does bot format R code cells in .ipynb files. Can we add support for this?

lionel-:
Sorry I closed this too fast. After researching the issue more, I found that:

1. With the CLI we could format notebooks tagged with R, or notebook cells tagged with R.

2. With the LSP, we could have specific support for `vscode-notebook` and `vscode-notebook-cell` URIs, as you have already noted in #432.

Relevant links from ruff:

- CLI dispatch for notebook
  https://github.com/astral-sh/ruff/blob/main/crates/ruff/src/commands/format.rs#L407-L424

- Server handler for notebook
  https://github.com/astral-sh/ruff/blob/6271fba1e1c01c9c49a360b1084217955145bce6/crates/ruff_server/src/server/api/requests/format.rs#L42-L53

- Notebook deserialiser and cell parser:
  https://github.com/astral-sh/ruff/blob/main/crates/ruff_notebook/src/notebook.rs

- Document data structure for a single cell:
  https://github.com/astral-sh/ruff/blob/6271fba1e1c01c9c49a360b1084217955145bce6/crates/ruff_server/src/edit/notebook.rs#L25-L29

- Extension URI selectors:
  https://github.com/astral-sh/ruff-vscode/blob/f0b48ed6e7bbfbdf41067a5471538d884184659f/src/common/utilities.ts#L44-L45

PR #432, opened 2025-10-21
Support R chunk formatting in Jupyter/Positron notebooks

kv9898:
Currently, R chunks in Jupyter/Positron `.ipynb` notebooks cannot be formatted by Air. I brought up this issue in #405 but it was closed as it was considered "out of scope for Air" for Air.  I opened another issue in https://github.com/issues/created?issue=posit-dev%7Cpositron%7C10059.

However, it turned out that by adding the following line to `lsp.ts` would enable Air to work for notebooks quite seamlessly:
```ts
{ language: "r", scheme: "vscode-notebook-cell" },
```

So I thought it would be worth writing a PR to enable this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # Development version
 
+- Added support for formatting notebook cells (progress towards #405, @kv9898).
+
+
 # 0.8.0
 
 - Added support for table formatting of `tribble()` and `fcase()` calls (#113).


lionel-:
Thanks a lot for taking a look at this!

It does work quite well but we're adding a newline at the end of chunks that shouldn't be there:

https://github.com/user-attachments/assets/dbfab65f-81e2-4aa4-9237-07cc4578f456

ruff explicitly trims this newline: https://github.com/astral-sh/ruff/blob/6271fba1e1c01c9c49a360b1084217955145bce6/crates/ruff_server/src/server/api/requests/format.rs#L121-L134

Would you like to also take a look at this issue? It would involve detecting cell URIs in the format request handler to trim the result before creating edits. I'm also happy to take it from here, just let me know.

kv9898:
@lionel- I also noted this, but i thought this was the desired behaviour. I would love to have a look but it will probably take me (or claude) a while to find out which part of the codebase is responsible for this.

Is there anything that stops us from mimicking ruff's solution?

BTW: i've also fixed the 2 PRs to ark.

kv9898:
@lionel- I think copilot did a pretty good job fixing the end-of-chunk empty line issue.

Note: if the empty line at then end is the *only* thing to be fixed, my current version would ignore it. But it won't add new empty lines, and, if there are other problems to be fixed, the trailing empty line would be removed.

PR #439, opened 2025-10-29
Add configuration guide for emacs

myoung3:


docs/editor-emacs.qmd
```diff
@@ -0,0 +1,62 @@
...
+# Using Eglot
+
+Add the following to your `init.el`:
+
+``` elisp
+(use-package eglot
+   :ensure t
+   :config
+   (require 'ess-site)
+   (add-hook 'ess-r-mode-hook 'eglot-ensure) ; start lsp in ess-r-mode
```
lionel-: Maybe `:after ess-site` would be more idiomatic with `use-package`

But probably we should provide raw configuration that people can adapt to their config manager, if they have any. I'm preparing some updates, I'll push them soon.

docs/editor-emacs.qmd
```diff
@@ -0,0 +1,62 @@
+---
+title: "Emacs"
+editor:
+  markdown:
+    wrap: sentence
+    canonical: true
+---
+
+# Using Eglot
```
lionel-: Should follow the structure of other editor docs

docs/editor-emacs.qmd
```diff
@@ -0,0 +1,88 @@
...
+## Format document or region
+
+- With `eglot`: Run `M-x eglot-format` to format the current buffer or selection.
+
+- With `lsp-mode`: Run `M-x lsp-format-buffer` to format the current buffer or `lsp-format-region` to format the current selection.
+
+
+# Known Issues
+
+- Attempting to format selection when multiple lines are selected seems not to be working, but selecting part or all a line in a multi-line expression will format that entire expression. This isssue occurs regardless of whether Air is used with Eglot or LSP-mode.
```
lionel-, line 112: Air is supposed to support format-region, so I'm a bit surprised about this. I just checked and it works in Positron / VS Code.

docs/editor-emacs.qmd
```diff
@@ -0,0 +1,88 @@
...
+
+## lsp-mode
+
+`lsp-mode` can be installed from MELPA, see https://emacs-lsp.github.io/lsp-mode/page/installation
+
+Adapt your `init.el` with the following configuration:
+
+```elisp
+;; Disable languageserver entirely
+(add-to-list 'lsp-disabled-clients 'lsp-r)
```
lionel-, line 38: Why is this needed? Because LSP-mode can't stack language servers? It would be good to add a note about this in this case.

docs/editor-emacs.qmd
myoung3: I verified this issue on a different OS with a clean Emacs installation. I can file a separate issue. I suggest leaving this warning in, but I'll defer to you.

docs/editor-emacs.qmd
myoung3: @lionel-  Please see the updated doc which includes an air-only config and a multi-lsp config for lsp-mode. Does this resolve the question?

docs/editor-emacs.qmd
myoung3: Lgtm

docs/editor-emacs.qmd
myoung3: Lgtm

docs/editor-emacs.qmd
```diff
@@ -0,0 +1,114 @@
...
+  "Formats buffer with Eglot on save for specified modes."
+  (when (member major-mode '(ess-r-mode))
+    (eglot-format-buffer)))
+
+(add-hook 'after-save-hook 'my-eglot-format-on-save))
+```
+
+With `lsp-mode`:
+
+```elisp
```
myoung3: lsp-mode actually has `lsp-format-buffer-on-save`. Would be better to use this:

> You can also trigger format on save by setting the variable lsp-format-buffer-on-save to a non-nil value. To select what major modes to format use lsp-format-buffer-on-save-list.

https://emacs-lsp.github.io/lsp-mode/page/main-features/

lionel-:
Thanks @myoung3!

PR #443, opened 2025-11-03
Converted from lspconfig to vim.lsp for Neovim 0.11 in docs

panda-stunseed:
The [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) documentation states that the require('lspconfig') is deprecated and will return a warning for Neovim 0.11+ users.

> require('lspconfig') (the legacy "framework" of nvim-lspconfig) https://github.com/neovim/nvim-lspconfig/issues/3693 in favor of [vim.lsp.config](https://neovim.io/doc/user/lsp.html#lsp-config) (Nvim 0.11+).

Since I had to do the conversion myself, I added another block in the documentation of the syntax for vim.lsp for >= 0.11+ users - if they wish to use it.

### DIFF
--- a/docs/editor-neovim.qmd
+++ b/docs/editor-neovim.qmd
@@ -24,7 +24,7 @@ lsp.config['air'] = {
         vim.api.nvim_create_autocmd("BufWritePre", {
             buffer = bufnr,
             callback = function()
-                lsp.buf.format()  
+                lsp.buf.format()
             end,
         })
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Should this be "the way" to use Air in Neovim then? 

Like, should this section come first and then we frame the other section as being for Neovim <0.11?

PR #450, opened 2025-11-19
Compute `ArgKind::Other` width from `text.chars().count()` not `text.len()`

### LINKED ISSUE #449: Incorrect table alignment with special characters
When a table contains some special characters (like `€` or `§`), the alignment isn't applied correctly:
```
tibble::tribble(
  ~col1          , ~col2 ,
  "A (Mio. €)" ,    10 ,
  "B (T €)"    ,    20 ,
  "C §"         ,    10 ,
  "works"        ,    10 ,
  "works too"    ,    10 ,
  "works!!"      ,    10 ,
  "$ works"      ,    10 ,
)
```

Tested on air version `0.8.0`

DavisVaughan:
Closes #449

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
 
 - Fixed an issue with Unicode elements and table alignment (#449).
 
-- Added support for formatting notebook cells (progress towards #405, @kv9898).
-
 
 # 0.8.0
 
--- a/docs/editor-neovim.qmd
+++ b/docs/editor-neovim.qmd
@@ -18,6 +18,28 @@ While the Air language server currently only supports formatting, it will gain m
 Add the following to your `init.lua`:
 
 ``` lua
+local lsp = vim.lsp
+lsp.config['air'] = {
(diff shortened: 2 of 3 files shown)


crates/air_r_formatter/src/r/auxiliary/call_arguments_table.rs
```diff
@@ -491,7 +491,13 @@ impl ArgKind {
...
                 None => *integer_width,
             },
-            ArgKind::Other { text } => text.len(),
+            ArgKind::Other { text } => {
+                // Unlike with integers and decimals, can't use `text.len()` here in case
+                // Unicode characters are involved. `len()` counts bytes, but we want to
+                // align based on the number of characters that visibly show up in a
+                // user's editor (#449).
+                text.chars().count()
+            }
```
DavisVaughan, line 500: We do use `text.len()` to compute `integer_width` in other paths, but that's fine, those can never have Unicode characters and using `len()` there will be a little faster I think.

Issue #269, opened 2025-02-27
Official Air `pre-commit` hook


DavisVaughan:
https://github.com/pre-commit/pre-commit

Pre-commit itself seems to be written in python, so requires python to install
https://pre-commit.com/#install

Ruff example: https://github.com/astral-sh/ruff-pre-commit

I think the vibe of how the Ruff one works is that they have this pre commit hook yaml:
https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml

Note the `ruff format --force-exclude` call.

And I think ruff itself gets installed from being listed as a requirement here:
https://github.com/astral-sh/ruff-pre-commit/blob/main/pyproject.toml

They do tag releases synced to each ruff version
https://github.com/astral-sh/ruff-pre-commit/tags

And this is how a user would reference that tag and call the hook:
https://github.com/astral-sh/ruff/blob/568cf88c6c5b5551a675ae2b13deedec0fe226cb/.pre-commit-config.yaml#L76-L79

lorenzwalthert:
You could also place the config file in the posit-dev/air repo directly. Then, if you push new tags to this repo, `pre-commit autoupdate` will recognize them. If you create a separate repo, you will have to push tags there as well and update the dependency on positron-dev/air every time you make a new release.

DavisVaughan:
Noting that what we really need is something like this https://github.com/pre-commit/pre-commit/issues/1453 to download pre installed binaries using pre-commit

DavisVaughan:
IIUC with *Unifying Approach* we'd set `language: rust` and pre-commit would `cargo install --bins` on the fly for our users (even bootstrapping rustup as required). While this is nice, I think we should not do that because we've gone through all the work of making easy to install binaries - we should use them if at all possible.

So that likely leaves us with *Splitting Approach*, where we need to decide which `language` to use. A few ideas:

## language: R

Create a pseudo-R package at `posit-dev/air-pre-commit`. The R package would be in charge of downloading the correct binary depending on the platform, and then running it. This would be pretty simple I think, we just need to map the OS and Architecture to the right GitHub Artifact URL

```
https://github.com/posit-dev/air/releases/download/<VERSION>/air-<ARCH>-<OS>.tar.gz
```

For `VERSION`, we'd put a constant in the package as `VERSION <- 0.4.1` which would correspond to the `tag` that we release at, so when people set `rev: 0.4.1` in their pre-commit-config.yml they get the right version.

[...]

PR #467, opened 2026-02-26
Add PyPI distribution workflows

DavisVaughan:
Closes https://github.com/posit-dev/air/pull/395
Lays the groundwork for https://github.com/posit-dev/air/issues/269

This PR adds GitHub Actions workflows for making Python wheels for air, which will be automatically uploaded to PyPI during our `release.yml` process managed by cargo-dist.

Importantly, we reuse the binaries already created in the `build.yml` step (including the signed Windows one).

I've called the Python package `air-formatter`. Unfortunately `air` is already taken, and PyPI yells at me that `air-r` is "too close" to something else that already exists.

---

Next step is to make `posit-dev/air-pre-commit`, which would look like https://github.com/astral-sh/ruff-pre-commit

You'd be able to do

```
repos:
- repo: https://github.com/posit-dev/air-pre-commit
  # Air version
  rev: v0.8.1
  hooks:
    # Run the formatter
    - id: air-format
```

allowing you and your team to pin the version of Air in your pre-commit hook.

Under the hood we'd use `language: python`, but users don't need to know that. It would look like

https://github.com/astral-sh/ruff-pre-commit/blob/3fe39287d165e090f89effa3985ca987300a6b6f/.pre-commit-hooks.yaml#L12-L21

and pre-commit would be in charge of installing a `air-formatter` wheel from PyPI and making `air` available as a CLI command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 # Development version
 
+- Air is now distributed on PyPI as `air-formatter` (#467).
+
+  This allows air to be invoked via uv:
+
+  ```bash
+  # Global install of `air`
+  uv tool install air-formatter
+  air format .
+
+  # One-off run
+  uvx --from air-formatter air format .
+  ```
+
 - Air is now code-signed on Windows (#461).
 
 


CONTRIBUTING.md
```diff
@@ -87,6 +109,20 @@ For a new release:
...
+
+Later on in `release.yml`, the `publish-pypi.yml` job runs at publish time. It collects the wheels and uses `uv publish` to send them off to PyPI. This is a specially named job! It uses PyPI's Trusted Publishing so that we don't need any tokens. Instead, on Davis's PyPI account we have told PyPI to expect that `posit-dev/air` has a `publish-pypi.yml` workflow with a `environment: pypi` GitHub Environment set up, and when binaries are pushed from that source, PyPI will accept them without any additional tokens.
+
+If you're testing the Python wheel generation locally, use `just build-wheel` to build the wheel, and `just run-wheel <air args>` to run it. This will build release Air, copy it into `scripts/`, build the "any" wheel (which is correct for you, since you just built Air), and then `run-wheel` will run it with `uv tool run`.
```
DavisVaughan, line 124: Documentation on how this works

Issue #202, opened 2025-01-23
Add support for stdin in `air format`


DavisVaughan:
Will need to scope out how exactly you specify that `air format` should be expecting stdin. Ruff does `ruff format -` and discards other file names when it sees that. It also has `--stdin-filename` which it uses for some things (not sure what, probably deciding if its an R file vs ipynb? We should take a look).

We will need to think about where we start to look for `air.toml`s from, if at all, in this case. I think ruff uses the current working directory, which I guess seems reasonable?

armish:
@DavisVaughan: following up on #385 -- I am happy to help with this if that won't be a bother for you and the team.

PR #471, opened 2026-03-04
Implement stdin support

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/202

New `--stdin-file-path <path>` option to enable stdin support in the Air CLI.

This is extremely similar to what most of the CLI formatting ecosystem does

```bash
biome --stdin-file-path <path>

prettier --stdin-filepath <path>

black --stdin-filename <path>

ruff format - --stdin-filename <path>
```

This option is a bit weird, and different formatters handle it slightly differently. 

## Three reasons to have `--stdin-file-path`

- It is the trigger to recognize that you're in "stdin mode". If it is supplied, we read from `stdin` and also error if any files or directories are provided the "normal" way.

- It determines where to start the search for `air.toml` files. This is extremely important!

    - For _preexisting_ files, if your IDE formats via stdin then we need to know where the stdin comes from. i.e. if stdin came from `project/file.R` and you have `project/air.toml`, we need `--stdin-file-path project/file.R` so we can find your config file

    - For _new_ files, where some tool has created R code but you haven't written it to disk yet, this is equally as important. If you want to write to `project/doesnt-exist-yet.R` and you have `project.air.toml`, we need `--stdin-file-path project/doesnt-exist-yet.R` to find your config file. Note that this means that the file provided _doesn't even have to exist yet_. And that's considered a feature!

- It determines the file extension to associate with stdin. For us, that is currently always `.R` and we don't really do anything with this. But for prettier, biome, and ruff, which can all format multiple file types, the `--stdin-file-path` tells them which formatter to use under the hood. Again, even if the file doesn't exist, the extension provides valuable info. I like that this leaves the door open for `.Rmd` or `.qmd` support if we want to add it in the future.

## Other miscellaneous details

- If a relative path is provided to `--stdin-file-path`, it is resolved from the current working directory. So basically `air format file.R` and `air format --stdin-file-path file.R` resolve the same way. I think this is right, but I would hope that any IDE that uses this feature would supply an absolute path as `--stdin-file-path` to avoid any funny business that could result from them launching air in a weird directory or something. Providing an absolute path removes any chance to miss an `air.toml`.

- I think our _default exclude_ behavior for `air format cpp11.R`, where `cpp11.R` is a `default-exclude`, is not quite right. I think that both here and in `air format --stdin-file-path cpp11.R`, we should _refuse_ to format these by default, and maybe provide some way to opt back in when you really want it. This would better support both pre-commit and RStudio, where they both blindly provide air whatever file the user saved or edited in, and they have no way of knowing what the excludes are. I think it is more important for that to refuse to format `cpp11.R` than it is for a user to explicitly run `air format cpp11.R` at the command line and get that file formatted. I'd argue we probably should not even provide an opt in flag to allow this at all. https://github.com/posit-dev/air/issues/472

- Similarly, I think our _default include_ behavior for `air format file.qmd`, where `file.qmd` is not in `default-include`, is not quite right. We've had a lot of people get confused that `air format file.qmd` fake works. If you don't get a parse error by chance, then air will just completely break your qmd. I'd again argue that we should _refuse_ to format this, maybe with a warning. Alternatively, if we feel like this goes a step too far, then I'd be ok with black listing `.qmd` and `.Rmd`, while still allowing everything else, like `air format NAMESPACE`, since that's technically an R file and isn't "excluded". https://github.com/posit-dev/air/issues/473

### DIFF
--- a/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_bash.snap
+++ b/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_bash.snap
@@ -72,12 +72,16 @@ _air() {
             return 0
             ;;
         air__format)
-            opts="-h --check --log-level --no-color --help [PATHS]..."
+            opts="-h --check --stdin-file-path --log-level --no-color --help [PATHS]..."
             if [[ ${cur} == -* || ${COMP_CWORD} -eq 2 ]] ; then
                 COMPREPLY=( $(compgen -W "${opts}" -- "${cur}") )
                 return 0
             fi
             case "${prev}" in
+                --stdin-file-path)
+                    COMPREPLY=($(compgen -f "${cur}"))
+                    return 0
+                    ;;
                 --log-level)
                     COMPREPLY=($(compgen -W "error warn info debug trace" -- "${cur}"))
                     return 0
--- a/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_elvish.snap
+++ b/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_elvish.snap
@@ -37,6 +37,7 @@ set edit:completion:arg-completer[air] = {|@words|
             cand help 'Print this message or the help of the given subcommand(s)'
         }
         &'air;format'= {
+            cand --stdin-file-path 'Use this option to enable reading from stdin and writing to stdout. This specifies a file path to associate the standard input with, which is used as the location to begin searching for configuration files from. The file does not have to exist and will not be read from. If a relative path is provided, it is resolved from the current working directory. If this option is specified, no other files or directories can be provided'
             cand --log-level 'The log level [default: warn]'
             cand --check 'If enabled, format results are not written back to the file. Instead, exit with a non-zero status code if any files would have been modified, and zero otherwise'
             cand --no-color 'Disable colored output. To turn colored output off, either set this option or set the environment variable `NO_COLOR` to any non-zero value'
--- a/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_fish.snap
+++ b/crates/air/tests/integration/snapshots/integration__generate_shell_completion__completions_fish.snap
@@ -43,6 +43,7 @@ complete -c air -n "__fish_air_needs_command" -f -a "format" -d 'Format a set of
 complete -c air -n "__fish_air_needs_command" -f -a "language-server" -d 'Start a language server'
 complete -c air -n "__fish_air_needs_command" -f -a "generate-shell-completion" -d 'Generate shell completion scripts'
 complete -c air -n "__fish_air_needs_command" -f -a "help" -d 'Print this message or the help of the given subcommand(s)'
+complete -c air -n "__fish_air_using_subcommand format" -l stdin-file-path -d 'Use this option to enable reading from stdin and writing to stdout. This specifies a file path to associate the standard input with, which is used as the location to begin searching for configuration files from. The file does not have to exist and will not be read from. If a relative path is provided, it is resolved from the current working directory. If this option is specified, no other files or directories can be provided' -r -F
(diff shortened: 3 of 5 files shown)


crates/air/src/commands/format/paths.rs
DavisVaughan, line 1: This was extracted from `format.rs` but nothing really changed, so you can skip review of this file

crates/air/src/commands/format/stdin.rs
```diff
@@ -0,0 +1,134 @@
...
+fn format_stdin(settings: &FormatSettings) -> Result<FormattedStdin, FormatStdinError> {
+    tracing::trace!("Formatting stdin");
+
+    let mut old = String::new();
+
+    // Blocks until EOF is received!
+    io::stdin()
+        .lock()
+        .read_to_string(&mut old)
+        .map_err(FormatStdinError::Read)?;
```
DavisVaughan, line 114: Air will "hang" if you just do this at the command line without providing it any stdin

```bash
air format --stdin-file-path foo.R
```

The process is waiting for you to write to stdin and will hang until you get an EOF.

I think this is all correct, it just caught me by surprise the first time

crates/air/tests/integration/format.rs
```diff
@@ -209,3 +210,356 @@ fn test_check_output_format() {
...
+            .arg(cpp11_path)
+            .run_with_stdin(cpp11_contents.to_string())
+    );
+
+    Ok(())
+}
+
+// TODO!: This should also refuse to format. See above.
+#[test]
+fn test_stdin_refuses_to_format_user_excludes() -> anyhow::Result<()> {
```
DavisVaughan, line 565: Two tests which I think should probably fail once we change the default `exclude` behavior, this would better support pre-commit and rstudio

crates/workspace/src/format.rs
```diff
@@ -1,36 +1,8 @@
...
 use air_r_formatter::context::RFormatOptions;
 use air_r_parser::Parse;
 use air_r_parser::RParserOptions;
 use thiserror::Error;
 
-use crate::settings::FormatSettings;
-
-/// Representation of a formatted file
-#[derive(Debug)]
-pub enum FormattedFile {
```
DavisVaughan, line 13: The little `format_file()` helper from here moved into `format/paths.rs` directly, and that allowed me to flatten a lot of unneeded abstraction and remove most of these intermediate enums

crates/air/tests/integration/format.rs
DavisVaughan: These now pass in https://github.com/posit-dev/air/pull/476

crates/air/tests/integration/format.rs
```diff
@@ -209,3 +210,393 @@ fn test_check_output_format() {
...
+    let air_path = "air.toml";
+    let air_contents = r#"
+[format]
+indent-width = 4
+"#;
+    std::fs::write(directory.join(air_path), air_contents)?;
+
+    // Running in `directory` with a relative path to `test_path`. No `air.toml` found in
+    // `directory` or its ancestors, so we use default settings.
+    insta::assert_snapshot!(
```
DavisVaughan, line 269: TODO: Fix this misleading test before merging

crates/air/src/commands/format/stdin.rs
lionel-: yep that's standard behaviour

crates/air/src/commands/format.rs
```diff
@@ -1,89 +1,43 @@
...
-                } else {
-                    Ok(ExitStatus::Failure)
-                }
-            } else {
-                Ok(ExitStatus::Error)
-            }
-        }
+fn check_argument_consistency(command: &FormatCommand) -> Option<ExitStatus> {
+    if command.stdin_file_path.is_some() && !command.paths.is_empty() {
+        tracing::error!(
```
lionel-, line 28: It's still feels weird to me that we communicate user-facing messages on stderr via the log infrastructure, even if it also makes sense 🙃

PR #501, opened 2026-06-10
Update codegen to use clippy approved let chains

DavisVaughan:
Without this, we get a ton of clippy warnings because we use 2024 edition Rust now. clippy and the Problems pane are now once again useful.

### DIFF
--- a/xtask/src/lib.rs
+++ b/xtask/src/lib.rs
@@ -47,7 +47,7 @@ pub fn reformat_without_preamble(text: impl Display) -> Result<String> {
     let _e = pushenv("RUSTUP_TOOLCHAIN", "stable");
     ensure_rustfmt()?;
     let output = run!(
-        "rustfmt --config newline_style=Unix";
+        "rustfmt --edition 2024 --config newline_style=Unix";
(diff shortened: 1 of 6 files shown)


xtask/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub fn reformat_without_preamble(text: impl Display) -> Result<String> {
     let _e = pushenv("RUSTUP_TOOLCHAIN", "stable");
     ensure_rustfmt()?;
     let output = run!(
-        "rustfmt --config newline_style=Unix";
+        "rustfmt --edition 2024 --config newline_style=Unix";
```
DavisVaughan, line 50: rustfmt defaults to 2015 edition! Let-chains did not exist there.

rustfmt does not respect rust-toolchain.toml or our top level `edition = "2024"` in `Cargo.toml` because it is meant to be used as a standalone tool

`cargo fmt` on the other hand would find and respect `edition = "2024"`, but we are using this thing to format stdin, which `cargo fmt` cannot do.

So after a bunch of research, I think hardcoding this is probably the best and easiest option

PR #502, opened 2026-06-10
Implement `assignment-style`

### LINKED ISSUE #359: Provide a way to standardize to `<-` or `=`
It would be great to be able to enforce the use of a consistent assignment operator (`<-` or `=`). Ideally with a default (likely `<-`?), and configurable in `air.toml`.

Thanks!

DavisVaughan:
Closes https://github.com/posit-dev/air/issues/359

- Defaults to `"preserve"`, no changes to assignment operators
- `"arrow"` changes all `=` to `<-`
- `"equal"` changes all `<-` to `=` _where possible_

The _where possible_ part is contained within `can_normalize_to_equal()`. It is really the only tricky part about this PR. I am fairly confident with it though. It is hand crafted after an extremely close read of `gram.y`, and double checked with a deep claude review.

The name `assignment-style` is meant to be similar to `indent-style` and an eventual `quote-style` (also used by ruff). The `"preserve"` name matches ruff's `quote-style = "preserve"`.

Note that I've chosen to make this opt-in. Hopefully people will find and use it. I considered setting `<-` as the default, but I was worried about aggressively changing people's code if they already use Air but prefer `=`.

In other news, let-chains are amazing

### DIFF
--- a/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
@@ -258,10 +258,6 @@ fn fmt_left_assignment_operator(
     f: &mut RFormatter,
 ) -> FormatResult<()> {
     match f.options().assignment_style() {
-        AssignmentStyle::Preserve => {
-            // Preserve as is
-            operator.format().fmt(f)
-        }
         AssignmentStyle::Arrow => match left_assignment_kind {
             LeftAssignmentKind::Equal => format_replaced(
                 operator,
@@ -290,6 +286,10 @@ fn fmt_left_assignment_operator(
                 }
             }
         },
+        AssignmentStyle::Preserve => {
+            // Preserve as is
+            operator.format().fmt(f)
+        }
     }
 }
 
--- a/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
@@ -6,14 +6,26 @@ use crate::r::auxiliary::call_arguments::FormatRCallArgumentsOptions;
 use air_r_syntax::AnyRExpression;
 use air_r_syntax::RBinaryExpression;
 use air_r_syntax::RBinaryExpressionFields;
+use air_r_syntax::RElseClause;
+use air_r_syntax::RExpressionList;
+use air_r_syntax::RForStatement;
+use air_r_syntax::RFunctionDefinition;
+use air_r_syntax::RIfStatement;
 use air_r_syntax::RLanguage;
+use air_r_syntax::RParenthesizedExpression;
+use air_r_syntax::RRepeatStatement;
 use air_r_syntax::RSyntaxKind;
+use air_r_syntax::RUnaryExpression;
+use air_r_syntax::RWhileStatement;
 use biome_formatter::FormatRuleWithOptions;
 use biome_formatter::format_args;
+use biome_formatter::prelude::dynamic_text;
+use biome_formatter::trivia::format_replaced;
 use biome_formatter::write;
 use biome_rowan::AstNode;
 use biome_rowan::SyntaxResult;
 use biome_rowan::SyntaxToken;
+use settings::AssignmentStyle;
 
 #[derive(Default, Debug, Clone, Copy)]
 pub(crate) enum ChainAlignment {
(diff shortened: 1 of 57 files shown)


crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -213,17 +225,221 @@ fn fmt_binary_assignment(
...
+    if RExpressionList::can_cast(parent_kind) {
+        return true;
+    }
+
+    // i.e. `(x <- 1)` to `(x = 1)`
+    if RParenthesizedExpression::can_cast(parent_kind) {
+        return true;
+    }
+
+    // i.e. `? x <- 1` to `? x = 1`
```
lionel-, line 357: I do wonder if we should leave `?` and `~` alone, since these are syntactically NSE.
No strong feeling either way though.

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -213,17 +225,221 @@ fn fmt_binary_assignment(
...
+    }
+
+    // i.e. `x <- y <- 1` to `x = y = 1`
+    // i.e. `x = y <- 1` to `x = y = 1`, but notably `x <- y = 1` is a parse error.
+    // i.e. `x <- 1 ? y` to `x = 1 ? y`
+    // i.e. `y ? x <- 1` to `y ? x = 1`
+    //
+    // We recurse through `can_normalize_to_equal()` in these cases because the parent
+    // binary expression must also be in a position where `=` would be valid. These would
+    // all result in parse errors or change semantic meaning if we didn't check this:
```
lionel-, line 372: 😬

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
```diff
@@ -213,17 +225,221 @@ fn fmt_binary_assignment(
...
+/// than `EQ_ASSIGN` (a superset, really). This is why we can always replace `=` with
+/// `<-`.
+/// https://github.com/wch/r-source/blob/b7e27523048d135e3a02560e51cb266702bd49c1/src/main/gram.y#L497
+///
+/// `EQ_ASSIGN` is also used in `sub:` and `formlist:`, but these correspond to named
+/// arguments (i.e. `fn(x = 1)`) and parameters with defaults (i.e. `function(x = 1) {}`)
+/// respectively, so are not relevant usages for us (and in fact those are locations where
+/// normalizing to `=` is not allowed).
+/// https://github.com/wch/r-source/blob/b7e27523048d135e3a02560e51cb266702bd49c1/src/main/gram.y#L549-L564
+fn can_normalize_to_equal(node: &RBinaryExpression) -> bool {
```
lionel-, line 336: A potentially more generic way of approaching this would be to only normalise assignment in statement expressions. Although I can't remember off the top of my head if we tend to consider braceless bodies as statements?

lionel- (approved):

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
DavisVaughan: We don't, which is why we have branches for if / for / while / etc with explicit examples about these exact cases

```
    // i.e. `if (cond) x <- 1` to `if (cond) x = 1`
    if let Some(parent) = RIfStatement::cast_ref(&parent)
        && let Ok(consequence) = parent.consequence()
        && consequence.syntax() == node
    {
        return true;
    }
```

```
    // i.e. `for(i in 1:5) x <- 1` to `for(i in 1:5) x = 1`
    if let Some(parent) = RForStatement::cast_ref(&parent)
        && let Ok(body) = parent.body()
        && body.syntax() == node
    {
        return true;
    }
```

```
    // i.e. `while(cond) x <- 1` to `while(cond) x = 1`
    if let Some(parent) = RWhileStatement::cast_ref(&parent)
        && let Ok(body) = parent.body()
        && body.syntax() == node
    {
        return true;
    }
```

For the most part, this PR is just normalizing in statement expression position. We detect when the parent is a `{}` or `()` and normalize in there. Then there's just a number of other less important cases we need to account for to make it work as you'd expect in things like braceless bodies. It's somewhat annoying that you have to account for them, but if you didn't then I'd definitely think Air had a bug.

crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
DavisVaughan: We explicitly handle `?` here, both one-sided and two-sided, and allow normalization in there. But that's obviously incredibly rare to see anyways so I'm not totally tied to it.

I guess with `~ (x <- 1)` we'd have to climb every parent of `<-` looking for `~` or `?` to see if we are nested within one of those before deciding to normalize to `=`?

---

Update: Actually, what you're saying here would also imply that there are some cases where `=` can't be converted to `<-`, which currently isn't true. Currently all `=` can be converted to `<-` and I kind of like how easy that is to explain. But you're saying `~ (x = 1)` shouldn't become `~ (x <- 1)`.

I think for that reason, I'd like to keep as is just so the `=` -> `<-` explanation is simple

DavisVaughan:
After some discussion with the tidyverse group, switching to `<-` as the default style.

- This is what all of us use, and we'd otherwise have to set `assignment-style = "arrow"` in our currently empty `air.toml`s to get this
- This is what the [style guide](https://style.tidyverse.org/syntax.html#assignment-1) already recommends
- This is the more flexible operator, as `=` can't be used in some places
- styler also does this by default
- The people who don't want this can still easily use preserve/equal.

PR #504, opened 2026-06-16
Prepend the active `air` executable location to the `PATH` of integrated terminals

### LINKED ISSUE #500: Positron does not put air on path
From the terminal:

```
hadleywickham@hadleywickham-MXP5 data-editor % which air
air not found
hadleywickham@hadleywickham-MXP5 data-editor % which quarto
/Applications/Positron.app/Contents/Resources/app/quarto/bin/quarto
```

DavisVaughan:
Closes #500 

Seems like this works pretty well. A few things to note are listed below.

- We turn off the default caching of `vscode.EnvironmentVariableCollection` with `persistent = false`. By default, it caches across workspace sessions so that if your extension set an environment variable once, then on the next open of Positron it can restore that environment variable tweak _before your extension even activates_. I worry about this, because it means that there can be a period of time where integrated terminals can try to use a "stale" path to an Air binary. Like if we've cached a path to Air, but then Positron updates and brings in a new version of Air, then the user opens Positron, then an integrated terminal that starts up will point to a nonexistent Air binary until the Air extension can start up and reset the env var to the correct updated value. It seemed better to me to just never cache, and only make the env var tweak after we've located the `binaryPath` and successfully started up an LSP client with it.

- Integrated terminals will only get an updated `PATH` if they are started _after_ the Air extension itself has activated. I think this is totally fine, as Air activates fairly aggressively if any R related content is found, and it would be pretty rare to go into a terminal and try to call `air` in a project that doesn't have any R files.

- I did a reasonable amount of work to ensure that we are dynamic to `air.addExecutableToPATH` configuration changes. i.e., if you turn it off, then if you immediately create a new terminal it should not have a `PATH` adjustment, no restart required. But note that we are purposefully not dynamic to `air.executableStrategy` or `air.executablePath`. The reason is that we always want to be in sync with _the running server_. So just changing those isn't enough to cause the running server to change, you also have to do `Air: Restart server` and then both the running server and the `PATH` adjustment will update together.

In the video you'll also see that your terminal turns yellow when an environment variable is out of sync! This is pretty cool!

https://github.com/user-attachments/assets/cc6c11b2-5185-491b-aa94-a5a570fc5ffe

Claude in the integrated terminal can find the bundled Air now:

Screenshot 2026-06-16 at 10 53 54 AM

### DIFF
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -103,7 +103,7 @@
 					"scope": "window",
 					"type": "string"
 				},
-				"air.addExecutableToPATH": {
+				"air.addExecutableToPath": {
 					"default": true,
 					"markdownDescription": "Whether to add the directory containing the `air` executable resolved by `air.executableStrategy` to the `PATH` of integrated terminals. This ensures that `air` invoked in an integrated terminal matches `air` used by the editor.",
 					"scope": "window",
--- a/editors/code/package.json
+++ b/editors/code/package.json
@@ -83,7 +83,7 @@
 				},
 				"air.executableStrategy": {
 					"default": "bundled",
-					"markdownDescription": "Strategy used to locate the `air` executable to start the language server with.",
+					"markdownDescription": "Strategy used to locate the `air` executable.",
 					"enum": [
 						"bundled",
 						"environment",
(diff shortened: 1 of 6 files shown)


editors/code/package.json
```diff
@@ -102,6 +102,12 @@
 					"markdownDescription": "The path to an `air` executable. Only utilized when `air.executableStrategy` is set to \"path\".",
 					"scope": "window",
 					"type": "string"
+				},
+				"air.addExecutableToPATH": {
```
DavisVaughan:
```suggestion
				"air.addExecutableToPath": {
```

PR #511, opened 2026-06-28
Upgrade dependencies to latest, including tree-sitter 0.26

Goldziher:
## Summary

Runs `cargo upgrade --incompatible` across the workspace and updates the lockfile, bringing every dependency to its latest release. The headline change is **tree-sitter 0.24 -> 0.26**.

### Motivation

I'm building a multi-linter that I'd like to use `air` as a library dependency within. The blocker was air's pin to an older `tree-sitter` (0.24): it conflicts with the `tree-sitter` version the rest of my toolchain requires, and Cargo cannot link two semver-incompatible copies of the `tree-sitter` runtime into one binary. Bumping air to `tree-sitter` 0.26 removes that hard blocker. While in there, I took the rest of the workspace's dependencies up to latest so the upgrade lands as one coherent change rather than leaving the tree partially stale.

### Notable bumps

- tree-sitter 0.24 -> 0.26
- clap / clap_complete 4.5 -> 4.6
- schemars 0.9 -> 1.2
- toml 0.8 -> 1.1, itertools 0.13 -> 0.15, strum 0.26 -> 0.28
- cargo_metadata 0.20 -> 0.23, similar-asserts 1.6 -> 2.0
- zed_extension_api 0.2 -> 0.7, bpaf 0.9.15 -> 0.9.26
- assorted patch-level bumps (anyhow, serde, tokio, tracing, uuid, and others)

### Fixups required by the upgrades

- Renamed the renamed clippy lint `empty_enum` -> `empty_enums` in the workspace lint table.
- Regenerated the shell-completion snapshots for clap_complete 4.6, which now namespaces internal subcommand identifiers with `__subcmd__`.
- Regenerated `air.schema.json` and its snapshot for schemars 1.2, which no longer inserts a leading space after embedded newlines in description strings. This is cosmetic; the schema semantics are unchanged.

### Validation

`cargo build --workspace` and `cargo clippy --workspace --all-targets` are clean. Snapshot tests were updated only where the upgrades changed generated output (completions and the JSON schema), and the changes were reviewed to be mechanical. The full cross-platform suite runs in CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aether_lsp_utils"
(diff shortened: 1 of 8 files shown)


DavisVaughan:
I appreciate the PR but I'm not comfortable bumping the (effectively "major") version of tree-sitter until tree-sitter-r is also bumped to 0.25 and then 0.26, handling all the details that fall out from that.
https://github.com/r-lib/tree-sitter-r/issues/187

tree-sitter ABI bumps are not a minor thing, and typically require a good chunk of time from me to look at, which is why I haven't gotten around to it yet.

I don't think I'll be able to merge this as is, so I'm going to close. I've had a few people ask about tree-sitter-r getting bumped to 0.26, so I'll try to prioritize that (but I'm also going on extended leave soon, so no promises)

PR #513, opened 2026-07-07
Implement user level `air.toml` support

### LINKED ISSUE #309: Consider a global fallback `air.toml` location
Right now, if we can't find an `air.toml` as we walk up the directory tree, then we give up and use the "default" settings of `Settings::default()`.

Ruff has support for also considering a user specified default configuration in this case, instead of `Settings::default()`:
https://docs.astral.sh/ruff/configuration/#config-file-discovery

> If no config file is found in the filesystem hierarchy, Ruff will fall back to using a default configuration. If a user-specific configuration file exists at ${config_dir}/ruff/pyproject.toml, that file will be used instead of the default configuration, with ${config_dir} being determined via [etcetera's native strategy](https://docs.rs/etcetera/latest/etcetera/#native-strategy), and all relative paths being again resolved relative to the current working directory.

I'm not 100% convinced we should spend time on this yet, but in #308 there is a reasonable case for wanting this so that all users / projects on an RStudio Server instance would get the same formatting out of the box.

Technically if you place an `air.toml` at root level `~/`, then I think we will find it for any project that doesn't have a "closer" `air.toml`, but that feels fairly unofficial and a little sketchy compared to a feature like this.

DavisVaughan:
Closes #309 

This PR adds user level `air.toml` support. It uses the etcetera crate, same as ark and ruff, to locate a config directory that is OS specific:

-   `~/.config/air/air.toml` on Linux and macOS. The environment variable `XDG_CONFIG_HOME` can be set as an alternative to `~/.config`, i.e. if set, Air will look in `{XDG_CONFIG_HOME}/air/air.toml`.

-   `%APPDATA%\air\air.toml` on Windows. The environment variable `APPDATA` can be set as an alternative to `%APPDATA%`, i.e. if set, Air will look in `{APPDATA}\air\air.toml`.

When there isn't a project level `air.toml`, Air will now look for a user level `air.toml` before falling back to the default Air settings. This makes the user level `air.toml` a useful place to put personal settings for scratch files or projects that don't use Air. We still highly encourage using a project level `air.toml` for collaborative projects.

### IDE Synchronization

A user level `air.toml` counts as a "real" `air.toml`, and takes precedence over IDE settings.

### Watched files notifications

We use `GlobPattern::Relative` to watch a directory _outside_ of the workspace, i.e.:

```rust
        watchers.push(FileSystemWatcher {
            glob_pattern: GlobPattern::Relative(RelativePattern {
                base_uri: OneOf::Right(directory.clone()),
                pattern: String::from("air.toml"),
            }),
            kind: None,
        });
```

where `directory` here is the user config directory of `~/.config/air/` (on mac). Note how we update `~/.config/air/air.toml` outside of the IDE and still get a change notification!

https://github.com/user-attachments/assets/1affb54c-3aa9-4353-a1cb-81fddb5df323

### `exclude`s

The most complicated part of this PR is `exclude` handling.

A user level `air.toml` lives _outside_ of your project's directory tree, so you can't specify _rooted_ (or relative) patterns like `R/foo.R` because this resolves to `{root}/R/foo.R`.

Instead, you must supply _unrooted_ (or global) patterns like `**/foo.R` or the simpler `foo.R`, which also resolves to `**/foo.R`. You can also do `folder/`, which resolves to `**/folder/`.

The simplest way to think about it is that you can't supply anything with _leading or interior `/`_, as that is relative to a `{root}`, which isn't applicable for user level `air.toml`s. I don't anticipate this being restrictive at all, and we give a pretty good error message about it. (These are also the same rules that git follows, as implemented by the ignore crate).

### DIFF
--- a/crates/air/src/commands/format/paths.rs
+++ b/crates/air/src/commands/format/paths.rs
@@ -15,6 +15,7 @@ use workspace::discovery;
 use workspace::discovery::DiscoveredSettings;
 use workspace::discovery::discover_r_file_paths;
 use workspace::discovery::discover_settings;
+use workspace::discovery::discover_user_settings;
 use workspace::format::FormatSourceError;
 use workspace::format::FormattedSource;
 use workspace::resolve::PathResolver;
@@ -38,7 +39,7 @@ pub(crate) fn format(
     exclude: discovery::Exclude,
     include: discovery::Include,
 ) -> anyhow::Result<ExitStatus> {
-    let mut resolver = PathResolver::new(Settings::default());
+    let mut resolver = PathResolver::new();
 
     for DiscoveredSettings {
         directory,
@@ -48,9 +49,11 @@ pub(crate) fn format(
         resolver.add(&directory, settings);
     }
 
+    let default_settings = discover_user_settings()?.unwrap_or_default();
+
     match mode {
         FormatMode::Write => {
-            let errors = format_paths_write(&paths, &resolver, exclude, include);
+            let errors = format_paths_write(&paths, &resolver, &default_settings, exclude, include);
 
             for error in &errors {
                 tracing::error!("{error}");
@@ -63,7 +66,8 @@ pub(crate) fn format(
             }
         }
         FormatMode::Check => {
-            let (paths, errors) = format_paths_check(&paths, &resolver, exclude, include);
+            let (paths, errors) =
+                format_paths_check(&paths, &resolver, &default_settings, exclude, include);
 
             for error in &errors {
                 tracing::error!("{error}");
@@ -98,16 +102,27 @@ fn inform_changed(paths: &[PathBuf], f: &mut impl Write) -> io::Result<()> {
 fn format_paths_write<P: AsRef<Path>>(
     paths: &[P],
     resolver: &PathResolver<Settings>,
+    default_settings: &Settings,
     exclude: discovery::Exclude,
     include: discovery::Include,
 ) -> Vec<FormatPathError> {
-    let paths = discover_r_file_paths(paths, resolver, discovery::Mode::Format, exclude, include);
+    let paths = discover_r_file_paths(
+        paths,
+        resolver,
+        default_settings,
+        discovery::Mode::Format,
+        exclude,
+        include,
+    );
 
     paths
         .into_iter()
         .filter_map(|path| match path {
             Ok(path) => {
-                let settings = resolver.resolve_or_fallback(&path);
+                let settings = resolver
+                    .resolve(&path)
+                    .map_or(default_settings, |item| item.value());
+
                 match format_path(&path, &settings.format) {
                     Ok(formatted) => match write_path(&path, formatted) {
                         Ok(()) => None,
@@ -124,16 +139,27 @@ fn format_paths_write<P: AsRef<Path>>(
 fn format_paths_check<P: AsRef<Path>>(
     paths: &[P],
     resolver: &PathResolver<Settings>,
+    default_settings: &Settings,
     exclude: discovery::Exclude,
     include: discovery::Include,
 ) -> (Vec<PathBuf>, Vec<FormatPathError>) {
-    let paths = discover_r_file_paths(paths, resolver, discovery::Mode::Format, exclude, include);
+    let paths = discover_r_file_paths(
+        paths,
+        resolver,
+        default_settings,
+        discovery::Mode::Format,
+        exclude,
+        include,
+    );
 
     paths
         .into_iter()
         .filter_map(|path| match path {
             Ok(path) => {
-                let settings = resolver.resolve_or_fallback(&path);
+                let settings = resolver
+                    .resolve(&path)
+                    .map_or(default_settings, |item| item.value());
+
                 match format_path(&path, &settings.format) {
                     Ok(file) => check_path(&path, file).map(Ok),
                     Err(err) => Some(Err(err)),
--- a/crates/air/tests/integration/config.rs
+++ b/crates/air/tests/integration/config.rs
@@ -1,0 +1,265 @@
+use std::path::Path;
+use std::path::PathBuf;
+use std::process::Command;
+use std::process::Stdio;
+
+use tempfile::TempDir;
+use workspace::config::user_config_directory_env_var;
+
+use crate::helpers::CommandExt;
+use crate::helpers::binary_path;
+
+fn write_user_air_toml(config: &Path, contents: &str) -> anyhow::Result<()> {
+    let directory = config.join("air");
+    std::fs::create_dir(&directory)?;
+    std::fs::write(directory.join("air.toml"), contents)?;
+    Ok(())
+}
+
+fn user_air_toml(config: &Path) -> PathBuf {
+    config.join("air").join("air.toml")
+}
+
+#[test]
+fn test_user_config_applies_when_no_project_air_toml() -> anyhow::Result<()> {
+    let user_config_directory = TempDir::new()?;
+    let user_config_directory = user_config_directory.path();
+    write_user_air_toml(user_config_directory, "[format]\nindent-width = 8\n")?;
+
+    let directory = TempDir::new()?;
+    let directory = directory.path();
+
+    let test_path = "test.R";
+    std::fs::write(directory.join(test_path), "if (TRUE) {\n1\n}\n")?;
+
+    // No project `air.toml`, so the user level `indent-width = 8` applies
+    let output = Command::new(binary_path())
+        .env(user_config_directory_env_var(), user_config_directory)
+        .current_dir(directory)
+        .arg("format")
+        .arg(test_path)
+        .run();
+
+    assert!(output.status.success());
+    assert_eq!(
+        std::fs::read_to_string(directory.join(test_path))?,
+        "if (TRUE) {\n        1\n}\n"
+    );
+
+    Ok(())
+}
+
+#[test]
+fn test_user_config_applies_to_stdin() -> anyhow::Result<()> {
+    let user_config_directory = TempDir::new()?;
+    let user_config_directory = user_config_directory.path();
+    write_user_air_toml(user_config_directory, "[format]\nindent-width = 8\n")?;
+
+    let directory = TempDir::new()?;
+    let directory = directory.path();
+
+    // The user level `indent-width = 8` applies to `--stdin-file-path` as well,
+    // and shows up in the snapshot
+    insta::assert_snapshot!(
+        Command::new(binary_path())
+            .env(user_config_directory_env_var(), user_config_directory)
+            .current_dir(directory)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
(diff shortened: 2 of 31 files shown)


crates/air/src/commands/format/paths.rs
```diff
@@ -48,9 +49,11 @@ pub(crate) fn format(
         resolver.add(&directory, settings);
     }
 
+    let default_settings = discover_user_settings()?.unwrap_or_default();
```
DavisVaughan, line 52: Really the only new thing in the CLI is that `Settings::default()` is replaced with this, where we first attempt to find user settings

crates/air/tests/integration/snapshots/integration__config__user_config_rooted_exclude_is_a_hard_error.snap
```diff
@@ -0,0 +1,18 @@
...
+success: false
+exit_code: 255
+----- stdout -----
+
+----- stderr -----
+air failed
+  Cause: Failed to parse [AIR_TOML]:
+Pattern `src/foo.R` must meet one of the following conditions:
+- Start with `**/`, i.e. `**/foo.R`
+- Contain no `/` characters unless it is at the very end, i.e. `foo.R` or `folder/`
```
DavisVaughan: This is the error you get when you have a bad `exclude`. Unfortunately I can't easily get `exclude` mentioned in there due to how the types are set up.

crates/lsp/src/capabilities.rs
```diff
@@ -32,13 +33,19 @@ impl AirClientCapabilities {
...
-            .and_then(|workspace| workspace.did_change_watched_files)
+            .and_then(|workspace| workspace.did_change_watched_files);
+
+        let dynamic_registration_for_did_change_watched_files = did_change_watched_files
             .and_then(|watched_files| watched_files.dynamic_registration)
             .unwrap_or_default();
 
+        let relative_pattern_support_for_did_change_watched_files = did_change_watched_files
+            .and_then(|watched_files| watched_files.relative_pattern_support)
+            .unwrap_or_default();
```
DavisVaughan, line 47: This is the Client side setting that determines if out-of-workspace watching is supported

crates/workspace/src/file_patterns.rs
```diff
@@ -6,42 +6,40 @@ use ignore::gitignore::Gitignore;
...
+/// When constructing the matcher, you supply a `root` path along with the `patterns` to
+/// be included in the matcher. When [RootedFilePatterns::matched] is called, the `root`
 /// path is always stripped from `path` before matching is done. This ensures that users
-/// can specify `"/special.R"` in their `air.toml` to match only `{root}/special.R`, and
-/// not also `{root}/subdir/special.R`.
+/// can specify `/special.R` in their `air.toml` to match only `{root}/special.R`, and not
+/// also `{root}/subdir/special.R`.
 #[derive(Clone, Debug)]
-pub struct FilePatterns {
+pub struct RootedFilePatterns {
```
DavisVaughan, line 17: We've renamed:
- `FilePatterns` -> `RootedFilePatterns`
- `DefaultFilePatterns` -> `UnrootedFilePatterns`

The "default" name made sense at the time because these were used in `DefaultExcludePatterns` and `DefaultIncludePatterns`, but it turns out this is the exact data structure we need for the general case of "I don't have a `root` path to resolve `patterns` relative to"

crates/workspace/src/resolve.rs
```diff
@@ -16,9 +16,6 @@ use std::path::{Path, PathBuf};
 /// See [`PathResolver::resolve()`] for more details on the implementation.
 #[derive(Debug, Default)]
 pub struct PathResolver<T> {
-    /// Fallback value to be used when a `path` isn't associated with any `items`
-    fallback: T,
-
```
DavisVaughan, line 21: `crates/lsp/src/workspaces.rs` got much simpler to implement after removing the `fallback` handling from `PathResolver` and instead relying on the caller to handle caller specific fallback details when `resolve()` returns `None`.

I was able to remove a few `// How can we do better?` style TODOs over there after clearing this up

crates/air/tests/integration/snapshots/integration__config__user_config_rooted_exclude_is_a_hard_error.snap
lionel-: I find this a bit confusing. Path separators are allowed in the pattern right? Which contradicts "unless it is at the very end"

crates/air/tests/integration/config.rs
```diff
@@ -0,0 +1,265 @@
...
+    );
+
+    Ok(())
+}
+
+#[test]
+fn test_project_air_toml_wins_over_user_config() -> anyhow::Result<()> {
+    let user_config_directory = TempDir::new()?;
+    let user_config_directory = user_config_directory.path();
+    write_user_air_toml(user_config_directory, "[format]\nindent-width = 8\n")?;
```
lionel-: Also add a second setting that overwrites the default, and check that the project does apply the default, not the user setting?

crates/lsp/src/handlers.rs
```diff
@@ -105,3 +92,49 @@ fn collect_regs(
...
+        glob_pattern: GlobPattern::String("**/air.toml".into()),
+        kind: None,
+    });
+    watchers.push(FileSystemWatcher {
+        glob_pattern: GlobPattern::String("**/.air.toml".into()),
+        kind: None,
+    });
+
+    // User configuration level `air.toml` lives outside any workspace folder, so we
+    // need `GlobPattern::Relative` if the IDE supports it
```
lionel-, line 111: Nice to put that capability (watching outside the workspace) to use!

crates/workspace/src/file_patterns.rs
```diff
@@ -156,15 +155,53 @@ impl DefaultFilePatterns {
     }
 }
 
+/// Check if a `pattern` is unrooted
+///
+/// Constructed by carefully analyzing [GitignoreBuilder]'s `add_line()`, which itself
+/// is derived from git's man page. It should be very stable.
```
lionel-, line 168: I'll trust you on that 😄

crates/workspace/src/file_patterns.rs
```diff
@@ -156,15 +155,53 @@ impl DefaultFilePatterns {
...
+        return true;
+    }
+
+    // Implicit `**/`, i.e.:
+    // - `foo.R`
+    // - `folder/`
+    // but not:
+    // - `foo/bar.R`
+    // - `/foo.R`
+    // - `foo/**/bar.R`
```
lionel-: This is kind of confusing (I know you didn't make the rules)

crates/workspace/src/file_patterns.rs
```diff
@@ -156,15 +155,53 @@ impl DefaultFilePatterns {
...
+    let pattern = pattern.strip_suffix('/').unwrap_or(pattern);
+    !pattern.contains('/')
+}
+
+fn err_rooted(pattern: &str) -> anyhow::Error {
+    anyhow::anyhow!(
+        "Pattern `{pattern}` must meet one of the following conditions:
+- Start with `**/`, i.e. `**/foo.R`
+- Contain no `/` characters unless it is at the very end, i.e. `foo.R` or `folder/`"
```
lionel-: Now I understand. But I think this might confuse our users, probably worth a note in docs

docs/configuration.qmd
```diff
@@ -72,6 +72,25 @@ If you run `air format` with a working directory of `~/packages/dplyr` or open y
...
+
+-   `~/.config/air/air.toml` on Linux and macOS.
+    The environment variable `XDG_CONFIG_HOME` can be set as an alternative to `~/.config`, i.e. if set, Air will look in `{XDG_CONFIG_HOME}/air/air.toml`.
+
+-   `%APPDATA%\air\air.toml` on Windows.
+    The environment variable `APPDATA` can be set as an alternative to `%APPDATA%`, i.e. if set, Air will look in `{APPDATA}\air\air.toml`.
+
+A user level `air.toml` takes precedence over IDE settings when [IDE settings synchronization](#configuration-settings-synchronization) is supported.
+
+Because user level `air.toml`s exist in a directory outside of your project's tree, any `exclude` patterns you provide must match at any depth.
```
lionel-:
```suggestion
Because user level `air.toml`s exist in a directory outside of your project's tree, any `exclude` patterns you provide must be _unrooted_, i.e. they must match at any depth.
```

docs/configuration.qmd
```diff
@@ -72,6 +72,25 @@ If you run `air format` with a working directory of `~/packages/dplyr` or open y
...
+Any pattern that begins with `**/` is allowed, as well as simple patterns such as `file.R` and `folder/`, which are internally treated as `**/file.R` and `**/folder/`.
```
lionel-:
```suggestion
An unrooted pattern is one that begins with `**/`. In addition, simple patterns such as `file.R` and `folder/`, are internally treated as `**/file.R` and `**/folder/`.
```

docs/configuration.qmd
```diff
@@ -72,6 +72,25 @@ If you run `air format` with a working directory of `~/packages/dplyr` or open y
...
+Patterns with a leading or interior `/` are not allowed, such as `/foo.R` and `R/foo.R`, as these imply `{root}/foo.R` and `{root}/R/foo.R`, which isn't applicable with a user level `air.toml`.
```
lionel-:
```suggestion
On the other hand, patterns with a leading `/` are _rooted_ and not allowed, e.g. `/foo.R`. In addition, patterns with an interior `/` like `R/foo.R` are also considered rooted and treated as `/R/foo.R`.

The distinction between rooted and non-rooted paths is a bit subtle but was chosen for consistency with Git rules.
```

crates/air/tests/integration/snapshots/integration__config__user_config_rooted_exclude_is_a_hard_error.snap
lionel-: Now I understand. Can we link to the doc here? Or clarify a bit by mentioning the notion of "rooted".

lionel- (approved):

crates/air/tests/integration/config.rs
DavisVaughan: good idea!

crates/air/tests/integration/snapshots/integration__config__user_config_rooted_exclude_is_a_hard_error.snap
DavisVaughan: Hopefully https://github.com/posit-dev/air/pull/513/changes/5b78a17ff39234a98a85ab6798a692601643c093 makes this all clearer with:
- consistent rooted vs unrooted terminology everywhere
- explicit "Unrooted patterns" section in the docs with bulleted comparison
- expanded error message information, which mirrors the "Unrooted patterns" docs (decided not to link to it, a lengthy error message describing the exact issue felt more useful in this case, and I'd hate to accidentally break the link one day in the future if we rework the site)

Issue #241, opened 2025-02-20
Epic: Formatting comments


lionel-:
Not planned in the short term but gathering ideas and concerns here:

- Roxygen (whole project in itself, will probably require a parser extension)

- Enforce space between comment sign and first character. Allow comment sign to be repeated and suffixed with a special character.

- Line length: 72 characters as in PEP8? https://peps.python.org/pep-0008/#maximum-line-length
  Prose is much easier to read with a small ligne length in general.

JosephBARBIERDARNAL:
Hey! Formatting for roxygen would be great. For example, I would love if the following would be possible:

before:

```r
#' @param path_config A character string specifying the file path to the YAML configuration file.
```

after:


```r
#' @param path_config
#' A character string specifying the file path to the YAML
#' configuration file.
```

or:


```r
#' @param path_config A character string specifying the file
#' path to the YAML configuration file.
```

DavisVaughan:
I have a particular style I like to use in a few packages where I include the argument _type_ in the documentation. This might be somewhat hard to auto format. I imagine there are a number of scenarios out there of people doing non standard ish documentation formatting, which doesn't necessarily look bad, but would generate a whole bunch of diffs if we do something like this. I'm not saying we shouldn't, it's just hard and subjective.

https://github.com/DavisVaughan/ivs/blob/a93a4467a6f15e6831d16908ae97f9adbd209e0e/R/iv.R#L99-L133

strengejacke:
[rewrap](https://open-vsx.org/extension/stkb/rewrap) is doing a fine job, including keeping indentions for lists etc. Just for inspiration...

emstruong:
This feature would be greatly appreciated.

PR #516, opened 2026-07-15
Implement roxygen2 `@examples` support

DavisVaughan:
Part of https://github.com/posit-dev/air/issues/241

The way this generally works is by taking over the default `fmt_leading_comments()` hook, replacing the default `biome_formatter::trivia::format_leading_comments()` helper that is used there with our own that can split leading comments into chunks of:
- standard comments handled by `biome_formatter::trivia::format_leading_comments()`
- roxygen2 blocks handled by our own custom formatter

Lots of design decisions for this one. I'll call them out one at a time below so we can have threaded conversations about each one. We should "resolve" each thread marked with **Important design decision:** before merging.

I've run this on dplyr, vctrs, rlang, purrr, and recipes and have refined based on the results and am pretty happy with the results.

https://github.com/user-attachments/assets/7a4ed537-5668-465d-80a7-12c553f13796

Hitting the user supplied line width, adjusted to account for `#' `:

https://github.com/user-attachments/assets/5e0f2666-424b-494b-935b-2cf19c8a323f


crates/air_r_formatter/src/trivia/roxygen/section/examples.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl Format<RFormatContext> for FormatRoxygenExamples<'_> {
+    fn fmt(&self, f: &mut RFormatter) -> FormatResult<()> {
+        if self.try_fmt(f)? {
+            return Ok(());
+        }
+
+        // Something failed, like a parse error. Totally fine and expected. Fall back to
+        // verbatim formatter.
+        FormatRoxygenVerbatim::new(self.section.comments(), self.prefix).fmt(f)
```
DavisVaughan, line 38: **Important design decision:**

We don't format `@examples` blocks with:
- parse errors
- `\dontrun{}` (or similar), which cause a parse error

We _silently_ swallow these and fall back to verbatim formatting.

crates/air_r_formatter/src/trivia/roxygen/section/examples.rs
```diff
@@ -0,0 +1,279 @@
...
+        // #' 2 + 2
+        // ```
+        //
+        // is normalized to
+        //
+        // ```
+        // #' @examples
+        // #' 1 + 1
+        // #' 2 + 2
+        // ```
```
DavisVaughan, line 73: **Important design decision:**

I think this makes sense. Should be rare, but I think it should be moved down.

crates/air_r_formatter/src/trivia/roxygen/section/examples.rs
```diff
@@ -0,0 +1,279 @@
...
+        // #' @param x A number.
+        // #'
+        // #' @examples
+        // #' fn(1)
+        // #' fn(2)
+        // #'
+        // #' @returns
+        // #' Something
+        // fn <- function(x) { x }
+        // ```
```
DavisVaughan, line 127: **Important design decision:**

Unlike a normal `.R` file, in an `@examples` block there are valid cases for wanting to keep at most 1 leading or trailing empty line. If the user had one in the original source, we keep it.

I think the trailing empty line is particularly compelling, and I tend to do that quite a bit.

crates/air_r_formatter/src/trivia/roxygen/section/examples.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Example sections always follow `#'`, i.e. they aren't the first thing on a
+        // line. This means they should always be indented with spaces, even if the user
+        // requests tabs for the surrounding file. But we do respect the user's
+        // `IndentWidth`.
+        let options = options
+            .clone()
+            .with_line_width(line_width)
+            .with_line_ending(LineEnding::Lf)
+            .with_indent_style(IndentStyle::Space)
+            .with_roxygen_examples(RoxygenExamples::Disabled);
```
DavisVaughan, line 174: When we format the `@examples` subdocument, we start with the user's options and override a few things:

- The `line_width` is reduced to account for `#' ` and any leading indents (think, nested R6 docs). This isn't an exact science, more on this later.
- We must use spaces for any indents. We aren't at the beginning of the line, we are past a `#'`, so emitting tabs at this point in time would be a bit crazy, and IDEs don't respect your "visible indent width" when you don't have tabs at the beginning of the line. We do still respect your indent-width size though.
- We force `\n` line breaks, which helps with splitting by `\n` later on.

crates/air_r_formatter/src/trivia/roxygen/section/examples.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        let text = printed.into_code();
+
+        // TODO: Can we make `InsertFinalNewline` an air option for library style usage?
+        // This isn't the first time we've had to strip it back off.
+        //
+        // Air currently strips off all user trailing line breaks and then unconditionally
+        // appends a single final `hard_line_break()` in `FormatRRoot`. We want to be in
+        // charge of trailing line breaks, so we strip that off here.
+        let text = text.strip_suffix('\n').unwrap_or(&text);
```
DavisVaughan, line 198: This isn't the first time I've had to strip off this unconditional `hard_line_break()`
https://github.com/posit-dev/air/blob/3f0d0062d26995f54312493ebf0b383b8bc35d65/crates/air_r_formatter/src/r/auxiliary/root.rs#L43

We also do it here
https://github.com/posit-dev/air/blob/3f0d0062d26995f54312493ebf0b383b8bc35d65/crates/lsp/src/handlers_format.rs#L155-L156

I think that it might be useful to make `InsertFinalNewline` an element of `FormatOptions` that isn't actually exposed to users, but can be used on the library side when embedding Air like we do here.

crates/air_r_formatter/src/trivia/roxygen/section/verbatim.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// Prints the content of the roxygen comment mostly verbatim, but:
+/// - Handles roxygen prefix normalization
+/// - Handles trailing whitespace trimming (same as `FormatLeadingComments`)
+///
+/// A simpler form of [biome_formatter::trivia::FormatLeadingComments]
+pub(crate) struct FormatRoxygenVerbatim<'a> {
+    comments: RoxygenComments<'a>,
+    prefix: &'a RoxygenPrefix,
+}
```
DavisVaughan, line 23: **Important design decision:**

For any non-`@examples` section, we fall back to a verbatim(ish) formatter.

The actual content of the roxygen comment line is printed verbatim, but we still do a little bit of normalization that feels correct to me:

- We trim trailing whitespace. This one is definitely correct. The original `FormatLeadingComments` did this too.

- We normalize the roxygen "prefix", i.e. either `#' ` or `##' `. We derive this prefix by taking whatever the _first_ line of the roxygen block is using, and normalize every line in that block to use the same prefix. As part of the normalization, we always ensure there is at least 1 space after the `'` token. i.e. `#'@param` is formatted as `#' @param`, even though we don't actually know how to format `@param` sections.

crates/air_r_formatter/src/trivia/roxygen/section/verbatim.rs
```diff
@@ -0,0 +1,67 @@
...
+            let range = comment
+                .piece()
+                .text_range()
+                .add_start(prefix_width)
+                .sub_end(suffix_width);
+
+            if !range.is_empty() {
+                // Write via zero copy `located_token_text()`
+                let token = comment.piece().as_piece().token();
+                write!(f, [space(), located_token_text(&token, range)])?;
```
DavisVaughan, line 57: This `located_token_text()` writer is pretty cool.

We could use `dynamic_text()` here and pass in the `&str` we want to format, which would be something like:

```
#' @param x This is a param
   |----------------------| # this is the `&str` we'd pass in 
```

And that would work fine except it would _clone_ the `@param x This is a param` because it needs to own the text for lifetime reasons.

Instead, we recognize that even if we apply the roxygen prefix and trailing whitespace normalization, the subset of the line that we want to write is present in the original source document, which means that we can refer to it by `range` and write it with `located_token_text()`. This does a _zero copy_ write, so it's more performant.

This is what comments are normally formatted with. To figure that out, look at biome's `Format<Context> for SyntaxTriviaPiece<L>` and then note how the use of `syntax_token_cow_slice()` there is using `FormatElement::LocatedTokenText` in the `Borrowed` case. We are basically doing the same thing here.

crates/air_r_formatter/src/trivia/roxygen/comments.rs
```diff
@@ -0,0 +1,215 @@
...
+/// Construction of [RoxygenSyntaxTriviaPieceComments] is guarded by [is_roxygen()],
+/// so we can expect a `#'` prefix
+#[inline]
+fn roxygen_strip_prefix(text: &str) -> &str {
+    let rest = text.trim_start_matches('#');
+    let rest = rest
+        .strip_prefix('\'')
+        .expect("`RoxygenSyntaxTriviaPieceComments` is guaranteed to have a `'`");
+    rest.strip_prefix(' ').unwrap_or(rest)
+}
```
DavisVaughan, line 177: The 3 helpers that basically power the whole PR 😄

crates/air_r_formatter/src/trivia/roxygen/format.rs
```diff
@@ -0,0 +1,129 @@
...
+            return Ok(());
+        };
+
+        // Derive roxygen prefix to normalize to from the first line of the block. We
+        // allow either `#'` or `##'`, with anything else being normalized to `#'`.
+        let prefix = match first.piece().hash_count() {
+            1 => RoxygenPrefix::Single,
+            2 => RoxygenPrefix::Double,
+            _ => RoxygenPrefix::Single,
+        };
```
DavisVaughan, line 37: **Important design decision:**

99% of us use `#'`, but `##'` is not uncommon when you search the GitHub mirror of CRAN. But there are effectively 0 hits of `###'`.

We print the roxygen prefix using a hardcoded zero copy `text("#'")` or `text("##'")` where that string has a `&' static str` lifetime, so it's very cheap and requires no allocations that way (see `RoxygenPrefix`).

But that means we don't support an arbitrary number of hashes. Which I think is fine. If we see some other number of hashes, we forcibly normalize to 1.

crates/air_r_formatter/src/trivia/roxygen/format.rs
```diff
@@ -0,0 +1,129 @@
...
+///   #' @examples
+///   #' fn(something_really_long_here)
+///   fn = function() {}
+/// )
+/// ```
+///
+/// Recomputing the adjusted line width now gives `80 - `#' `.len() - 2 = 75`, and on a
+/// second pass that would cause a 76 character wide `fn(something_really_long_here)` to
+/// break. This goes against idempotence, but is such a rare case that we don't worry
+/// about it.
```
DavisVaughan, line 88: **Important design decision:**

To format the `@examples` section, we compute an _adjusted line width_ based on the user's preferred line width, minus the `#' ` prefix, minus any leading indent _in the original source_ (think, an R6 class with docs that are indented).

99.9% of the time this is totally fine, but there are extremely rare cases where this can result in idempotence issues for 1 pass, and then it stabilizes. I think this section documents this limitation quite well, and we have explicit tests for it. I'm cool with it.

crates/air_r_formatter/src/trivia/roxygen/section.rs
```diff
@@ -0,0 +1,266 @@
...
+pub(crate) enum FormatRoxygenSection<'a> {
+    /// Leading description before any roxygen tags
+    Introduction(verbatim::FormatRoxygenVerbatim<'a>),
+
+    /// `@examples` and `@examplesIf`
+    Examples(examples::FormatRoxygenExamples<'a>),
+
+    /// An unhandled roxygen tag
+    Unknown(verbatim::FormatRoxygenVerbatim<'a>),
+}
```
DavisVaughan, line 28: We parse the entire roxygen block into _sections_ by _tag_. Each section is printed independently with its own rules. Unknown sections are printed verbatim. This sets us up to add _prose_ formatting in the future as well.

This:

```r
#' The introduction
#'
#' Some more introduction lines
#' 
#' @param x Some stuff
#'
#' @param y Some stuff that is really long
#'   and overflows onto the next line
#' 
#' @examples
#' 1 + 1
#' 2 + 2
```

breaks into these sections:

```r
#' The introduction
#'
#' Some more introduction lines
#' 
```

```r
#' @param x Some stuff
#'
```

```r
#' @param y Some stuff that is really long
#'   and overflows onto the next line
#' 
```

```r
#' @examples
#' 1 + 1
#' 2 + 2
```

crates/air_r_formatter/src/trivia/leading_comments.rs
```diff
@@ -0,0 +1,50 @@
...
+use crate::trivia::roxygen::RoxygenComments;
+
+/// Air's replacement for [biome_formatter::trivia::format_leading_comments()]
+///
+/// Splits a `node`'s leading comments into runs of:
+/// - [crate::trivia::roxygen::FormatRoxygenComments]
+/// - [biome_formatter::trivia::FormatLeadingComments]
+pub(crate) fn format_leading_comments(node: &RSyntaxNode) -> FormatLeadingComments<'_> {
+    FormatLeadingComments { node }
+}
```
DavisVaughan, line 15: Our main entry point that replaces `biome_formatter::trivia::format_leading_comments()`. This formats both standard leading comments and recognized roxygen comments.

crates/air_r_formatter/src/lib.rs
```diff
@@ -230,10 +231,13 @@ where
...
-    /// inside of the `fmt_fields` method or customize the formatting of the leading comments.
+    /// You may want to override this method if you want to manually handle the formatting
+    /// of comments inside of the `fmt_fields` method or customize the formatting of the
+    /// leading comments, but note that in Air we have special handling of leading
+    /// roxygen2 comments, so keep that in mind if you plan to override this.
     fn fmt_leading_comments(&self, node: &N, f: &mut RFormatter) -> FormatResult<()> {
-        format_leading_comments(node.syntax()).fmt(f)
+        // Our method, not `biome_formatter::trivia::format_leading_comments()`!
+        crate::trivia::format_leading_comments(node.syntax()).fmt(f)
     }
```
DavisVaughan, line 241: Here's where we override the _default_ handling of all leading comments that Air sees

crates/workspace/src/toml_options.rs
```diff
@@ -123,6 +124,23 @@ pub struct FormatTomlOptions {
...
+    /// roxygen2 block and the leading `#'` comment characters.
+    ///
+    /// Parse failures within an example section are silent and do not affect the parse
+    /// status of the containing file.
+    ///
+    /// Rd markup, like `\dontrun{}` and `\donttest{}`, are not supported and will result
+    /// in the entire `@examples` or `@examplesIf` section being left unformatted.
+    ///
+    /// This option is disabled by default.
+    pub roxygen_examples: Option<bool>,
```
DavisVaughan, line 142: **Important design decision:**

This option is currently called `roxygen-examples` and it is _opt in_.

Given that:
- we already do some basic normalization to non-`@examples` sections in this PR
- we probably want to attempt to format the prose at some point

I'm thinking that a better name for this option is probably

```
roxygen = true / false
```

Which is probabllllly still `false` for now as we shake out any bugs (?), and you can opt in with `true`.

And in the future if we add prose formatting then we may want to expand the option's value set to

```
roxygen = false # completely off
roxygen = true # opt in to examples and prose formatting
roxygen = "examples" # only examples, no prose

# this doesn't make any sense to me. if you are already using Air and want us to touch
# your roxygen in any way, then you almost certainly want us to touch your examples
roxygen = "prose"
```

PR #520, opened 2026-07-22
fix: preserve walrus comments across formatter passes

### LINKED ISSUE #512: Air isn't idempotent when `:=` expression is broken over multiple lines
Starting from this code:

```r
y :=
  # comment
  a
```
After the first `air format .`:

```r
y := # comment
a
```
After the second `air format .`:

```r
y := a # comment
```

Tested with Air 0.10.0

msaule:
Fixes #512.

## Summary

Preserve an own-line comment between `:=` and its RHS so that formatting is idempotent. Comment-free walrus line breaks and end-of-line comments retain Air's existing compact canonical output.

## Root cause

Walrus formatting intentionally flattens ordinary line breaks. When the RHS instead starts with an own-line comment, flattening changes the comment's syntactic position; a second pass then moves it after the RHS.

## Validation

- `cargo test -p air_r_formatter`
- `cargo test --workspace --exclude lsp`
- `cargo clippy -p air_r_formatter --all-targets`
- compiled CLI two-pass checks for own-line, comment-free, and end-of-line comment cases

### DIFF
--- a/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
+++ b/crates/air_r_formatter/src/r/auxiliary/binary_expression.rs
@@ -161,7 +161,9 @@ fn fmt_binary_sticky(
 /// Walrus assignment is not considered when looking for persistent line breaks because we
 /// don't want the `:=` case below to look like a request for expansion. While `:=` is
 /// technically parsed as a binary operator, we format it more like a named argument with
-/// a simple `space()` between the operator and the right hand side.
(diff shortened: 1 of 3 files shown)


lionel-:
This probably needs a bit more thinking about how this fits in with other binary operators, interacts with persistent line breaks, etc. Davis is on parental leave since last week and I'm about to take mine next week, so unfortunately this will need to wait until Fall. Thanks a lot for taking a look at this though!

PR #525, opened 2026-09-22
Include rowwiseDT as a default table-formatted call

MichaelChirico:
`data.table::rowwiseDT` is analogous to `tribble()`, some examples:

https://github.com/PsyTeachR/webexercises/blob/be725898ae910100e67d08577ca54e247d746f20/revdep/library/pema/data.table/doc/datatable-joins.R#L37-L44

https://github.com/jcprentice/epimodels/blob/f8cc8ec097360d9548a440ecd3db5ac25c9543a7/R/make_parameters.R#L327-L338

https://github.com/tdhock/2024-08-ift603-712/blob/30f257e435ca242cb65e74e486a1f05dd8bed294/cv-subsets-small-large-noise-fr.R#L187-L191


lionel-:
It does look like it!

Could you please add a test so that we keep an eye on that use case.

MichaelChirico:
Good call! The fact that the table 'header' uses missing arguments tripped up the original implementation, so it winds up being a bit more complicated. Tests and documentation updated.

lionel-:
oh hmm I see.

Does it mean trailing empty arguments should close the table formatting then? Could you add more tests for this case?

